PR #33, opened 2017-10-13
 Only the chevron should expand the child on click when the menuItem has a link

Xstoudi:


nnixaa:
Hi @Xstoudi, thanks for your PR. Frankly, I'm not sure we should allow such behavior at all, as this is at least not obvious, plus such change means that we have two possible behaviors in one places which is unacceptable from UX standpoint. What do you think?

Xstoudi:
Hi @nnixaa 
Look at this screenshot:
capture du 2017-10-13 16-54-24

As you can see, I've to projects that are already created and when I click on "Projects" it should open the Projects management (delete, create, edit).

I don't see a better way to do it and the suggested behavior doesn't seem to be so bad to me.

nnixaa:
Well yeah, it is obvious when you have only one behavior in your project at the same time. But what if you have both? Say projects are expanded only but a click on the 'chevron' icon, and another menu item, for instance, Clients, which doesn't have a category landing page (so clients page has no link) with "delete create edit", but has these pages as children like this:
Clients
 - List
 - Create new
 - Stats

In this case, you will have two menu items _looking exactly the same_ but behaving differently. And this is the rule we cannot break - one thing can only act in one way.
So in my view, we have to think of a better way both covering your needs and not confusing anyone :)

Xstoudi:
I am open to any proposal to solve this. Until this, I'll keep the fork to answer my needs.

nnixaa:
Closing as we will redo the menu.

PR #34, opened 2017-10-14
Make it possible to set the href link.

caciocode:


nnixaa:
Hey @caciocode thanks for your PR. Could you please complete it with the following points:

- lint checks are passing
- tests are added/updated
- showcase in the playground updated
- SCSS variables added/updated // not needed in this case
- tsdocs added/updated
- commit message is properly formatted
- for the override - registered in a list of overrides  // not needed in this case
- component *.theme registered in a list of component themes  // not needed in this case
- looks great on all default themes  // not needed in this case

PR #78, opened 2017-11-16
Let logout.component use form configs

lionax:
This will logout.component use the form configs in the same way as the login.component does.


nnixaa:
Hi @lionax, many thanks! I added a couple of changes and we are ready to merge.

PR #81, opened 2017-11-19
fix(height): menu small max-height

hryb:


nnixaa:
@hryb I believe we need to calculate this value dynamically based on items count multiplied by item height.

PR #90, opened 2017-11-24
feat(auth-provider): firebase auth provider

MolochkoVitaly:


lolsborn:
I'm trying to use this code as I would like to use Firebase with nebular, but I keep getting `firebase.auth() is not a function` any ideas?

tanwarsatya:
@lolsborn  were you able to resolve this.

lolsborn:
I have not.  I didn't spend any time on it this weekend.  Even though the
code I share is probably enough to reproduce it I was thinking about
creating a minimum project that someone can checkout which reproduces the
error.  If that sounds useful let me know.

--
Steven Osborn
http://lolsborn.com

Kshitij-Banerjee:
The code works. I just copied parts of it into my project and its good (y) thanks!

src/framework/auth/providers/firebase-auth.provider.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Firebase registration.
+   *
+   * @param data any
+   * @returns Observable<NbAuthResult>
+   */
+  register(data?: any): Observable<NbAuthResult> {
+    return Observable.fromPromise(firebase.auth().createUserWithEmailAndPassword(data.email, data.password))
+      .map((res) => {
```
Kshitij-Banerjee: should be flatMap? 
otherwise the result is Observable<NbAuthResult| Obsevable<NbAuthResult>>

src/framework/auth/providers/firebase-auth.provider.ts
```diff
@@ -0,0 +1,220 @@
...
+        return this.processSuccess(res, this.getConfigValue('logout.redirect.success'),  []);
+      })
+      .catch((res) => {
+        return Observable.of(this.processFailure(res, this.getConfigValue('logout.redirect.failure'),
+          [res.message]));
+      });
+  }
+
+  private processSuccess(response?: any, redirect?: any, messages?: any): NbAuthResult {
+    return new NbAuthResult(true, response, redirect, [], messages);
```
Kshitij-Banerjee: should add the last parameter rawToken, otherwise tokenService ignores this payload and does not store it.

Kshitij-Banerjee: infact, since this doesn't return a JWT. we need to store `JSON.stringify(response)` as the rawToken

Kshitij-Banerjee:
@nnixaa  Thanks !  
Also, please do check the token storage mechanisms (coz I was getting issues with it). After some fixes it works, but i still have to manually call tokenService.clear() on logout() for some reason.

PR #105, opened 2017-12-19
Fix typo in request password sub-title

thecodershome:


nnixaa:
Thanks for noticing, @thecodershome.

PR #145, opened 2018-01-17
Fix getActionEndpoint when null

Nightbr:

For example for logout, if you set logout: { endpoint: null, }, but you have a baseEndpoint, it will execute DELETE {baseEndpoint}/null on your api. This will fix the thing.


nnixaa:
Hi @Nightbr, thanks for you PR! We are in process of creating tests for auth so will be able to accept changes after we complete the tests.

nnixaa:
Hi @Nightbr, we now have unit tests for the auth service, do you think you can add a test for this case?

Nightbr:
Hey, I will try to add a test for this case.
Is all auth tests are in the file https://github.com/akveo/nebular/blob/master/src/framework/auth/services/auth.spec.ts?

Nightbr:
I think this test already test the logout:
https://github.com/akveo/nebular/blob/master/src/framework/auth/services/auth.spec.ts#L294
I just add 3 lines to return null when actionEndpoint is null and don't return 'baseApiUrl/undefined'.
I don't think this need another test to the base test.
Let me know what do you think.

eeskildsen:
This would be useful. I'm directing it to a dummy endpoint in the meantime.

nnixaa:
Thanks @Nightbr, closing in favor of https://github.com/akveo/nebular/pull/1211

PR #309, opened 2018-03-20
refactor(auth): pass providers through special `install` method for b…

nnixaa:
…etter type check

BREAKING CHANGES:
Auth module options' `providers` list changed from key=>value object to array.
So now instead of
```
...
providers: {
  email: {
    service: NbEmailPassAuthProvider,
    config: { ... }
  }
}
```
we do this:

```
providers: [
  NbEmailPassAuthProvider.install('email', { ... }),
]
```
This way we can type-check the configuration object for each provider specified.

### DIFF
--- a/docs/articles/auth-install.md
+++ b/docs/articles/auth-install.md
@@ -8,7 +8,7 @@
 ## Installation steps
 
 1) First, let's install the module as it's distributed as an npm package, but make sure you have the [Nebular Theme module up and running](https://akveo.github.io/nebular/#/docs/installation/add-into-existing-project).
-Nebular Theme is required to use built-in Auth Components. If you are not going to use those at all, you can use `Auth Module` without `Nebular Theme`. 
+Nebular Theme is required to use built-in Auth Components. If you are not going to use those at all, you can use `Auth Module` without the `Nebular Theme` module. 
 
 Let's assume that we need to setup email & password authentication based on Nebular Auth and NbEmailPassAuthProvider.
 
@@ -18,7 +18,7 @@ Let's assume that we need to setup email & password authentication based on Nebu
 
 `import { NbEmailPassAuthProvider, NbAuthModule } from '@nebular/auth';`
 
-3) Now, let's configure the module by specifying available providers, in your case we add `NbEmailPassAuthProvider`:
+3) Now, let's configure the module by specifying available providers, in our case we add `NbEmailPassAuthProvider`:
 
 ```typescript
 
@@ -28,7 +28,7 @@ Let's assume that we need to setup email & password authentication based on Nebu
     
    NbAuthModule.forRoot({
      providers: [
-       NbEmailPassAuthProvider.install('email', { ... }),
+       NbEmailPassAuthProvider.register('email', { ... }),
      ],
      forms: {},
    }), 
@@ -37,7 +37,8 @@ Let's assume that we need to setup email & password authentication based on Nebu
 
 ```
 
-We also specified a `forms` key, which configures available options for the Auth Components. We'll leave it empty for now and get back to it in the [Configuring UI](#/docs/auth/configuring-ui) article.
+We also specified a `forms` key, which configures available options for the Auth Components. 
+We leave it empty for now and get back to it in the [Configuring UI](#/docs/auth/configuring-ui) article.
 
 4) Next, we need to configure Auth Components routes, let's add those into your `app-routing.module.ts`:
 
--- a/docs/articles/auth-install.md
+++ b/docs/articles/auth-install.md
@@ -27,16 +27,11 @@ Let's assume that we need to setup email & password authentication based on Nebu
    // ...
     
    NbAuthModule.forRoot({
-         providers: {
-           email: {
-             service: NbEmailPassAuthProvider,
-             config: {
-              ...
-             },
-           },
-         },
-         forms: {},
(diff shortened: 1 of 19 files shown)


src/framework/auth/services/token/token-packer.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { NB_AUTH_TOKENS } from '../../auth.options';
+import { NbAuthSimpleToken, NbAuthToken, nbCreateToken } from './token';
+
+interface NbTokenPack {
+  class: string,
+  value: string,
+}
+
+/**
```
Tibing: I think you miss description here.

src/framework/auth/services/token/token-packer.ts
```diff
@@ -0,0 +1,44 @@
...
+      value: token.toString(),
+    });
+  }
+
+  unpack(rawToken: string): NbAuthToken {
+    let tokenPack: NbTokenPack;
+    try {
+      tokenPack = JSON.parse(rawToken);
+    } catch (e) {
+      console.warn('NbAuthToken: Can not unpack token, using empty `NbAuthSimpleToken`.');
```
Tibing: Maybe we have to throw an error and stop *unpack* execution? I don't think it's correct to warn and return an empty token.

src/framework/auth/services/token/token-packer.ts
```diff
@@ -0,0 +1,44 @@
+import { Inject, Injectable } from '@angular/core';
+
+import { NB_AUTH_TOKENS } from '../../auth.options';
+import { NbAuthSimpleToken, NbAuthToken, nbCreateToken } from './token';
+
+interface NbTokenPack {
```
Tibing: Also, I don't like this naming at all. When I see *token pack* I think about a number of tokens. Not about information required to recreate token.  Please, think about it and choose more informative naming for this feature.

docs/articles/auth-install.md
```diff
@@ -27,22 +27,18 @@ Let's assume that we need to setup email & password authentication based on Nebu
...
-             service: NbEmailPassAuthProvider,
-             config: {
-              ...
-             },
-           },
-         },
-         forms: {},
-       }), 
+     providers: [
+       NbEmailPassAuthProvider.register('email', { ... }),
```
Tibing: Do you really think we have to duplication **email** in the name of the provider and in the first param of the *register* function? Maybe we can remove this duplication somehow?

docs/articles/auth-provider.md
```diff
@@ -24,60 +24,62 @@ We assume you already have the Auth module installed inside of your `*.module.ts
...
+     providers: [
+       NbEmailPassAuthProvider.register('email', { ... }),
+     ],
+   }),
   ],
 });
 
 ```
 
+`email` here is an alias we assign to the provider, so that we can mention it later on dynamically. Plus we can configure multiple providers with various configurations
```
Tibing: Why don't we just use *NbEmailPassAuthProvider* in this case?

src/framework/auth/providers/dummy-auth.provider.ts
```diff
@@ -63,7 +70,11 @@ export class NbDummyAuthProvider extends NbAbstractAuthProvider {
         ['Something went wrong.']);
     }
 
-    // TODO is it missed messages here, is it token should be defined
-    return new NbAuthResult(true, this.createSuccessResponse(data), '/', ['Successfully logged in.']);
+    return new NbAuthResult(true,
```
Tibing: Why don't we just pass an object in this constructor instead of the multiple params?
I think it would be much prettier.

src/framework/auth/services/token/token-storage.ts
```diff
@@ -27,27 +27,28 @@ export abstract class NbTokenStorage {
...
+    super();
   }
 
   /**
    * Returns token from localStorage
    * @returns {NbAuthToken}
    */
   get(): NbAuthToken {
-    return nbCreateToken(this.tokenClass, localStorage.getItem(this.key));
+    return this.tokenPacker.unpack(localStorage.getItem(this.key));
```
Tibing: Maybe, it would be cleaner to not call one function in another? And do this operation in two lines?

src/framework/auth/services/token/token-storage.ts
```diff
@@ -27,27 +27,28 @@ export abstract class NbTokenStorage {
...
   }
 
   /**
    * Sets token to localStorage
    * @param {NbAuthToken} token
    */
   set(token: NbAuthToken) {
-    localStorage.setItem(this.key, token.toString());
+    localStorage.setItem(this.key, this.tokenPacker.pack(token));
```
Tibing: This to

src/framework/auth/services/token/token-packer.ts
nnixaa: Let's add a token status (empty, invalid, valid, expired) or in the next version think about throwing an error.

src/framework/auth/providers/dummy-auth.provider.ts
nnixaa: Check todo

src/framework/auth/services/token/token-packer.ts
nnixaa: Parcel.

Issue #270, opened 2018-02-23
Hidden options on menu should be fixed


gemins:
### Issue type

**I'm submitting a ...**  (check one with "x")

* [X] bug report
* [ ] feature request

### Issue description

**Current behavior:**
The html file on menu components not have hidden on the correct tag, then when a some menu is hidden, the container steal having the border lines. (The tag without *ngIf function is "li.menu-item").
Please check that, thanks you.

**Expected behavior:**
The Menu should be hidden entire.

**Steps to reproduce:**
Set hidden option on some menu to true.

PR #312, opened 2018-03-22
fix(menu): avoid DOM elements creation for hidden menu items (fix#270)

mishkolesnikov:
Fix for issue #270


nnixaa:
@mishkolesnikov let's add a test for this case also

nnixaa:
@mishkolesnikov could you check the conflicts?

PR #383, opened 2018-04-23
Feat/multiple tokens

nnixaa:

### DIFF
--- a/src/framework/auth/auth.module.ts
+++ b/src/framework/auth/auth.module.ts
@@ -33,13 +33,15 @@ import { routes } from './auth.routes';
 import { deepExtend } from './helpers';
 
 export function nbStrategiesFactory(options: NbAuthOptions, injector: Injector): NbAuthStrategy[] {
-  return options.strategies
-    .map(([strategyClass, strategyOptions]: [NbAuthStrategyClass, NbAuthStrategyOptions]) => {
+  const strategies = [];
+  options.strategies
+    .forEach(([strategyClass, strategyOptions]: [NbAuthStrategyClass, NbAuthStrategyOptions]) => {
       const strategy: NbAuthStrategy = injector.get(strategyClass);
       strategy.setOptions(strategyOptions);
 
-      return strategy;
+      strategies.push(strategy);
     });
+  return strategies;
 }
 
 export function nbOptionsFactory(options) {
--- a/src/framework/auth/auth.module.ts
+++ b/src/framework/auth/auth.module.ts
@@ -6,15 +6,23 @@ import { HttpClientModule } from '@angular/common/http';
 
 import { NbCardModule, NbCheckboxModule, NbLayoutModule } from '@nebular/theme';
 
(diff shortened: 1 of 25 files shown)


src/framework/auth/auth.module.ts
```diff
@@ -33,13 +41,24 @@ import { routes } from './auth.routes';
 import { deepExtend } from './helpers';
 
 export function nbStrategiesFactory(options: NbAuthOptions, injector: Injector): NbAuthStrategy[] {
-  return options.strategies
-    .map(([strategyClass, strategyOptions]: [NbAuthStrategyClass, NbAuthStrategyOptions]) => {
+  const strategies = [];
+  options.strategies
+    .forEach(([strategyClass, strategyOptions]: [NbAuthStrategyClass, NbAuthStrategyOptions]) => {
```
Tibing: Why do you use forEach here?

src/framework/auth/auth.module.ts
```diff
@@ -33,13 +41,24 @@ import { routes } from './auth.routes';
...
-      return strategy;
+      strategies.push(strategy);
+    });
+  return strategies;
+}
+
+export function nbTokensFactory(strategies: NbAuthStrategy[]): NbAuthTokenClass[] {
+  const tokens = [];
+  strategies
+    .forEach((strategy: NbAuthStrategy) => {
```
Tibing: And here

src/framework/auth/services/token/token.ts
```diff
@@ -1,25 +1,33 @@
...
+
+  getName(): string {
+    return (this.constructor as NbAuthTokenClass).NAME;
+  }
 }
 
-export interface NbTokenClass {
-  new (raw: string): NbAuthToken
+export interface NbAuthTokenClass {
+  NAME: string;
```
Tibing: I don't think **NAME** is a good name for this para. Maybe it has to be a sort of id?

src/framework/auth/auth.module.ts
```diff
@@ -33,13 +41,24 @@ import { routes } from './auth.routes';
...
       const strategy: NbAuthStrategy = injector.get(strategyClass);
       strategy.setOptions(strategyOptions);
 
-      return strategy;
+      strategies.push(strategy);
+    });
+  return strategies;
+}
+
+export function nbTokensFactory(strategies: NbAuthStrategy[]): NbAuthTokenClass[] {
```
Tibing: Also, It would be great if you add tokens validation. We have to warn a user about missed and same tokens.

src/framework/auth/services/token/token.ts
nnixaa: it's fine I suppose, as token.getId() would be confusing, expecting identity id rather than id of token type.

src/framework/auth/auth.module.ts
nnixaa: actutlly it could be the case, when two strategies use the same token with same `NAME`

PR #397, opened 2018-04-26
Feat/oaut2

nnixaa:


src/framework/auth/services/token/token.ts
```diff
@@ -121,3 +125,93 @@ export class NbAuthJWTToken extends NbAuthSimpleToken {
     return super.isValid() && (!this.getTokenExpDate() || new Date() < this.getTokenExpDate());
   }
 }
+
+const validateOAuth2Token = (data) => {
```
Tibing: why validation parse data?

src/framework/auth/services/token/token.ts
```diff
@@ -121,3 +125,93 @@ export class NbAuthJWTToken extends NbAuthSimpleToken {
...
+  constructor(protected data: { [key: string]: string|number }|string = {}) {
+    // we may get it as string when retrieving from a storage
+    super(validateOAuth2Token(data));
+  }
+
+  /**
+   * Returns the token value
+   * @returns string
+   */
+  getValue(): string {
```
Tibing: I don't think the naming of the *getValue* and *getRefresh* is okay. Maybe, it would be better name this methods: *getAccessToken* and *getRefreshToken*?

src/framework/auth/services/token/token.ts
```diff
@@ -121,3 +125,93 @@ export class NbAuthJWTToken extends NbAuthSimpleToken {
...
+    }
+
+    const date = new Date();
+    date.setUTCSeconds(new Date().getUTCSeconds() + Number(this.token.expires_in));
+
+    return date;
+  }
+
+  /**
+   * Validate value and convert to string, if value is not valid return empty string
```
Tibing: I think you've missed token validation 😄 
Frankly speaking, I don't think it's a good idea to validate token and return empty string if it's invalid.
Lets keet method as it is and just fix the comment.

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -0,0 +1,321 @@
...
+          if (!result) {
+            this.authorizeRedirect();
+            return observableOf(null);
+          }
+          return this.getAuthorizationResult();
+        }),
+      );
+  }
+
+  getAuthorizationResult(): Observable<any> {
```
Tibing: As I understand correctly there will be more types: code, token, password, credentials. So, let's create a map with type -> redirect result instead of *isType** functions and ugly if. What do you think?

src/framework/auth/services/token/token.ts
nnixaa: so `token.getAccessToken`? 🤓

Tibing: I think it would be better than *getValue()*

PR #440, opened 2018-05-23
refactor(examples): use common patter for each example

Tibing:

- rename example -> stacked-example
- add example names
- refactor example components to fit updated examples model

### DIFF
--- a/src/framework/theme/components/popover/popover.directive.ts
+++ b/src/framework/theme/components/popover/popover.directive.ts
@@ -28,7 +28,7 @@ import { NbPlacementHelper } from './helpers/placement.helper';
 /**
  * Powerful popover directive, which provides the best UX for your users.
  *
- * @example(popover/popover-showcase.component)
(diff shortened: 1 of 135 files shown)


src/framework/theme/components/popover/popover.directive.ts
```diff
@@ -77,7 +77,7 @@ import { NbPlacementHelper } from './helpers/placement.helper';
  * - Hover mode works like hint mode with one exception - when the user moves mouse from host element to
  * the container element popover will not be hidden.
  *
- * @example(popover/popover-modes.component)
+ * @stacked-example(Available modes, popover/popover-modes.component.html)
```
nnixaa: why not rename everywhere?

Tibing: Because we'll have to add names for all examples and I think it would slightly convenient to rename in that time.

Tibing: done

PR #536, opened 2018-06-29
Oauth2 Password Grant-type implementation

alain-charles:
### **oAuth2 password grant-type implementation**
 
This pull request implements the **password grant type** oAuth2 Strategy authentication (this was a Todo)
This implementation of oAuth2 password authentication uses standards of the [RFC 6749](https://tools.ietf.org/html/rfc6749#page-19).

Request:
`grant_type=password&username=johndoe&password=A3ddj3w`

Response
```
{
       "access_token":"2YotnFZFEjr1zCsicMWpAA",
       "token_type":"example",
       "expires_in":3600,
       "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA",
 }
```

**App.js has been updated** for handling this new oAuth2 Grant-type 
Playground has been added the **/oauth2-password route for demo**

I still have to implement the corresponding spec.ts
I plan to implement the possibility to support a basic authentication against the authorization server when using oAuth2 password grant-type (client id/client secret in the request header)

My next pull request will be about jwt-interceptor (automatic refreshtoken sending when accesstoken has expired)

Hope you will merge.

Sincerely yours
Alain


src/backend/app.js
```diff
@@ -64,6 +64,43 @@ app.post('/api/auth/login', function (req, res) {
...
+    var user = users.find(function (u) {
+      return u.email === email && u.password === password;
+    });
+    if (user) {
+      var payload = {
+        id: user.id,
+        email: user.email,
+        role: 'user',
+      };
+      var refreshPayload = {
```
nnixaa: why do we need an additional payload for a refresh token?

src/framework/auth/strategies/oauth2/oauth2-strategy.options.ts
```diff
@@ -39,6 +40,15 @@ export class NbOAuth2AuthStrategyOptions {
     endpoint: 'authorize',
     responseType: NbOAuth2ResponseType.CODE,
   };
+  password?: {
```
nnixaa: why do we need to intorduce an additional section of parameters for password grant type? 
As far as I can tell it is basically a token request but with a different grant_type (https://tools.ietf.org/html/rfc6749#section-4.3.2). So we can still use `token: {}` key passing these settings like this:
```
token: {
  grantType: NbOAuth2GrantType.PASSWORD,
} 
```

what do you think?

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -84,7 +84,7 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
   }
 
   get responseType() {
-    return this.getOption('authorize.responseType');
+      return this.getOption('authorize.responseType');
```
nnixaa: unnecessary spaces

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -154,17 +154,22 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
...
-          if (!result) {
-            this.authorizeRedirect();
-            return observableOf(new NbAuthResult(true));
-          }
-          return this.getAuthorizationResult();
-        }),
-      );
+  authenticate(data?: any): Observable<NbAuthResult> {
+
+    if (this.getOption('password') && data) {
```
nnixaa: there is no need to check on the `data` I believe. Probably we better check  it like this:
```
if (this.getOption('token.grantType') === NbOAuth2GrantType.PASSWORD) {
} else {
 //
}
```

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -211,6 +216,39 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
...
+            true,
+            res,
+            this.getOption('redirect.success'),
+            [],
+            this.getOption('defaultMessages'),
+            this.createToken(res));
+        }),
+        catchError((res) => {
+          const errors = [];
+          if (res instanceof HttpErrorResponse) {
```
nnixaa: let's make it like this
```
if (res instanceof HttpErrorResponse) {
            errors = this.getOption('defaultErrors');
          } else {
            errors.push('Something went wrong.');
          }
```
in this case `Something went wrong` message indicates that this is not an HTTP error, so most likely an internal error happened on a client side.

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -280,6 +318,19 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
...
   }
 
+  protected buildPasswordRequestData(email: string, password: string ): any {
+    const params = {
+      grant_type: this.getOption('password.grantType'),
+      email: email,
+      password: password,
+    };
+
+    Object.entries(params)
```
nnixaa: let's move this part
```
Object.entries(params)
      .forEach(([key, val]) => !val && delete params[key]);```

into some `cleanParams` function, as this block is duplicated three times now :)

File: src/playground/oauth2-password/oauth2-password-login.component.ts [hunk 7]
```diff
@@ -0,0 +1,146 @@
+/**
+ * @license
+ * Copyright Akveo. All Rights Reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+import {Component, Inject, OnDestroy} from '@angular/core';
+import { NbAuthOAuth2Token, NbAuthResult, NbAuthService, NB_AUTH_OPTIONS } from '@nebular/auth';
+import { Router } from '@angular/router';
+import {getDeepFromObject} from '../../framework/auth/helpers';
```
nnixaa: spaces

src/playground/oauth2-password/oauth2-password-login.component.ts
```diff
@@ -0,0 +1,146 @@
+/**
+ * @license
+ * Copyright Akveo. All Rights Reserved.
+ * Licensed under the MIT License. See License.txt in the project root for license information.
+ */
+
+import {Component, Inject, OnDestroy} from '@angular/core';
```
nnixaa: spaces

src/playground/oauth2-password/oauth2-password-login.component.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  ngOnDestroy(): void {
+  }
+
+  getConfigValue(key: string): any {
+    return getDeepFromObject(this.options, key, null);
+  }
+
+  getClaims(rawToken: string): string {
+    return this.jwtHelper.decodeToken(rawToken);
```
nnixaa: I don't think we need a new library to handle token decoding, even though this is a demo part, not framework.
Stil,l we can use our `NbAuthJWTToken` to decode token value of `NbAuthOAuth2Token`:

```
nbAuthCreateToken(NbAuthJWTToken, rawToken).getPayload();
```

src/framework/auth/strategies/oauth2/oauth2-strategy.options.ts
alain-charles: You are right. 
I'm  going to simplify this immediately

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
alain-charles: right

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
alain-charles: Test:wp failed as i was not checking data. 
Your proposition is great, going to try.

src/backend/app.js
alain-charles: I don't understand what you mean.
In the response we have to send an access_token and a refresh_token so i need 2 payloads for 2 tokens?
What do you think ?

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
alain-charles: OK for pushing "Something went wrong" in case of error being internal and not HTTP.
But i would like to return the error message contained in the 401 http response, as described in RFC.

So i propose something like this:
```
if (res instanceof HttpErrorResponse) {
            if (res.error.error_description) {
              errors.push(res.error.error_description);
            } else {
              errors = this.getOption('defaultErrors');
            }
          } else {
            errors.push('Something went wrong.');
          }
```

nnixaa: Yes, I agree, that perfectly makes sense. The only thing is that we probably need to modify the other method to conform with this style (so that all other requests will firstly look to errors into the responses, then in the config and only then report `something went wrong`.
So that it makes sense to move the error extraction into the additional function, doesn't it?

nnixaa:
@alain-charles great work btw!

src/playground/oauth2-password/oauth2-password-login.component.ts
alain-charles: done

src/backend/app.js
alain-charles: You are right, the specification does not impose anything about the refresh token.
I have seen some implementations of refresh_token considering it can be a JWT refreshToken
But it does not have to.
So you're right, i will fall back to basic string.

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
alain-charles: gonna push lines below
```
  protected handleResponseError(res: any): Observable<NbAuthResult> {
    let errors = [];
    if (res instanceof HttpErrorResponse) {
      if (res.error.error_description) {
        errors.push(res.error.error_description);
      } else {
        errors = this.getOption('defaultErrors');
      }
    } else {
      errors.push('Something went wrong.');
    }
    return observableOf(
      new NbAuthResult(
        false,
        res,
        this.getOption('redirect.failure'),
        errors,
        [],
      ));
  }
```

alain-charles:
@nnixaa Latest commit on pull request takes into account all that we discussed together.
Please let me know if you request other changes.

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -192,21 +206,27 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
             this.createToken(res));
         }),
         catchError((res) => {
-          let errors = [];
-          if (res instanceof HttpErrorResponse) {
-            errors = this.getOption('defaultErrors');
-          } else {
-            errors.push('Something went wrong.');
-          }
+          return this.handleResponseError(res);
```
nnixaa: let's simplify this a bit 
`catchError((res) =>this.handleResponseError(res))`

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -192,21 +206,27 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
...
+        map((res) => {
+          return new NbAuthResult(
+            true,
+            res,
+            this.getOption('redirect.success'),
+            [],
+            this.getOption('defaultMessages'),
+            this.createToken(res));
+        }),
+        catchError((res) => {
```
nnixaa: let's simplify this a bit 
`catchError((res) =>this.handleResponseError(res))`

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -234,21 +254,7 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
...
-
-          return observableOf(
-            new NbAuthResult(
-              false,
-              res,
-              this.getOption('redirect.failure'),
-              errors,
-              [],
-            ));
+          return this.handleResponseError(res);
```
nnixaa: let's simplify this a bit 
`catchError((res) =>this.handleResponseError(res))`

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -260,10 +266,7 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
       redirect_uri: this.getOption('token.redirectUri'),
       client_id: this.getOption('clientId'),
     };
-
-    Object.entries(params)
-      .forEach(([key, val]) => !val && delete params[key]);
-
+    this.cleanParams(params);
```
nnixaa: `return this.cleanParams(params);`

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -273,13 +276,41 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
       refresh_token: token.getRefreshToken(),
       scope: this.getOption('refresh.scope'),
     };
+    this.cleanParams(params);
```
nnixaa: `return this.cleanParams(params);`

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -273,13 +276,41 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
...
-    Object.entries(params)
-      .forEach(([key, val]) => !val && delete params[key]);
-
+  protected buildPasswordRequestData(email: string, password: string ): any {
+    const params = {
+      grant_type: this.getOption('token.grantType'),
+      email: email,
+      password: password,
+    };
+    this.cleanParams(params);
```
nnixaa: `return this.cleanParams(params);`

src/playground/oauth2-password/oauth2-password-login.component.ts
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+
+  logout() {
+    this.authService.logout('password')
+      .subscribe((authResult: NbAuthResult) => {
+      });
+  }
+
+  ngOnDestroy(): void {
```
nnixaa: let's remove an empty method

package-lock.json
```diff
@@ -11,7 +11,7 @@
       "dev": true,
```
nnixaa: revert it please since there are no changes to `package.json`

alain-charles:
@nnixaa I have difficulties reverting package-lock.json. I do not really know how to resync my fork and the upstream repo. I tried but it was not successful.
Maybe you can help me or maybe you can exclude it from merge.

nnixaa:
@alain-charles It seems to me you already excluded it.

nnixaa:
I guess we are good to merge it!

nnixaa:
@alain-charles sorry, last thing, could you add the new grant type here to the comment block https://github.com/akveo/nebular/pull/536/files#diff-93149eac5909aab260e620aeff722cdaR35? so that it gets inserted into the documentation.

alain-charles:
@nnixaa i pushed the line you were waiting in the comment block.

nnixaa:
@alain-charles and it's done! Congrats on your first contribution to Nebular 👍

alain-charles:
I do think it is a wonderful project. Here we will be users of it, so if i can contribute it is great.

I'll post very soon a new PR concerning jwtTokenInterceptor. What i want (automatic refresh expired token) works with our backend here, but i have still  to develop the backend in app.js for the playground.

PR #560, opened 2018-07-13
Refresh token requests fully managed by NbAuthService & NbAuthJWTInterceptor

alain-charles:
### Refresh token requests fully managed by NbAuthService & NbAuthJWTInterceptor

This pull requests intends to provide the necessary code so that refreshtoken requests are automatically sent to the backend when the local stored accesstoken has expired.
This allows the user not to type his login/password as long as his refresh-token is valid.
The refreshtoken requests are send transparently.

### Run it in the playground
You can easily test the code by calling the url http://localhost:4200/#/auth/call-api.
Then the AuthGuard service redirects you to the login page.

The strategy used to log in is Oauth2 with password grant-type, so you can use example@akveo.com.

You are logged with an access token whose ttl is 60 seconds and a refresh token whose ttl is 120 seconds.
These values can be modified in the configuration of the backend (config.js).

capture d ecran 2018-07-13 a 08 15 51

### Call of the secured api
Once logged, go immediately to http://localhost:4200/#/auth/call-api and click the call API button.

The XHR request is send with the authorization header containing Bearer + the access-token.
You can see that Alain's got good wines in his cellar.

capture d ecran 2018-07-13 a 08 17 05

### Automatic refresh-token request
Wait a minute, and click once again on the call-api button.

Observe the network logs of you browser : the acces-token was expired, so the framework sent automatically a refresh request to the backend.
capture d ecran 2018-07-13 a 08 18 52

It got a new access token, used in the call-api request.
capture d ecran 2018-07-13 a 08 19 20

### Automatic redirect to login
If you wait once a minute, and if you click the call-api button, you are redirected to the login page, because refresh-token itself expired.

capture d ecran 2018-07-13 a 08 19 39

### Used principles
Originally, NbAuthService was not injected any strategy name used for login.
This implies that you have to give the strategy name when calling its refreshToken method.

I decided not to change this, and i had to store the strategy name used for login with the token itself (this makes sense, the token was obtained using a strategy name).
I also had to store in the token the expiration date because Oauth2 sends a expires-in property.

I modified unitary tests.

Hope you will like this work.

### DIFF
--- a/docs/app/app.module.ts
+++ b/docs/app/app.module.ts
@@ -40,7 +40,7 @@ const docs = require('../output.json');
     NbCheckboxModule,
     NbProgressBarModule,
     NbMenuModule.forRoot(),
-    NbThemeModule.forRoot({ name: '' }),
+    NbThemeModule.forRoot({ name: 'default' }),
     NgdThemeModule.forRoot(),
     NbSidebarModule.forRoot(),
     RouterModule.forRoot(routes, { useHash: false }),
--- a/src/framework/auth/services/auth.spec.ts
+++ b/src/framework/auth/services/auth.spec.ts
@@ -24,12 +24,13 @@ describe('auth-service', () => {
   let tokenService: NbTokenService;
   let dummyAuthStrategy: NbDummyAuthStrategy;
   const testTokenValue = 'test-token';
+  const ownerStrategyName = 'strategy';
 
   const resp401 = new HttpResponse<Object>({body: {}, status: 401});
   const resp200 = new HttpResponse<Object>({body: {}, status: 200});
 
-  const testToken = nbAuthCreateToken(NbAuthSimpleToken, testTokenValue);
-  const emptyToken = nbAuthCreateToken(NbAuthSimpleToken, null);
+  const testToken = nbAuthCreateToken(NbAuthSimpleToken, testTokenValue, ownerStrategyName);
+  const emptyToken = nbAuthCreateToken(NbAuthSimpleToken, null, ownerStrategyName);
 
   const failResult = new NbAuthResult(false,
     resp401,
--- a/src/framework/auth/services/auth.spec.ts
+++ b/src/framework/auth/services/auth.spec.ts
@@ -26,6 +26,7 @@ describe('auth-service', () => {
   const testTokenValue = 'test-token';
   const ownerStrategyName = 'strategy';
 
+
   const resp401 = new HttpResponse<Object>({body: {}, status: 401});
   const resp200 = new HttpResponse<Object>({body: {}, status: 200});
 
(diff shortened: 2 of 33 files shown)


nnixaa:
@alain-charles thanks again for the PR! I will review it shortly.

nnixaa:
Hey @alain-charles, so I walked through the files and I must say you did a huge work. However, I don't think we could merge it like this as:
- the changes are quite deep in terms of architecture and it's hard to predict the impact on the other features and possible regression 
- the PR introduces a lot of breaking changes which we are really trying to avoid

My suggestion would be to break this PR into smaller steps starting from very basic and then moving to a particular functionality this PR is aiming to add. I see it the following way, but this is a subject for a discussion:

- so the first thing I noticed is that currently storing a token we cannot tell which strategy is responsible for it, so basically we don't have a back-link to the strategy. To fix this we need to store the `owner strategy` name. Not sure at the moment where else this could be useful, but for the refresh token we definitely should be able to recognize "whom to ask to refresh the token".

- the second thing is a fix for the `expires_in` calculation. Currently, you implemented it the way we store the exact `expDate`. I would slightly adjust it so that we store no the exact date, but the `createdAt` date. This it gives us a bit more flexibility,  e.g. if in the future we have a similar case when we need to tell when a token is created to calculate something there will be no need to add an additional value.

- .. and then we can revise the approach to the refresh token related changes

What do you think? I understand this adds some work for you which might not be very interesting for you, but from the other side, we are at the stage when we should be very careful merging some huge changes.

alain-charles:
I agree this is a huge PR but my goal was the automatic refreshtoken request to be sent so I needed all I did .

I agree we can insert the changes in multiple PRs. The is no problem for me. It is more work for me but here we want to stay synchronized with nebular so I will do the necessary.

Do you want me to send a PR with point 1 and 2 above ? Or only point 1?
Do you agree with the way the strategy is stored in the code I propose ?

Please let me know what I can concretely do,

Alain

nnixaa:
> I agree we can insert the changes in multiple PRs. The is no problem for me. It is more work for me but here we want to stay synchronized with nebular so I will do the necessary.

Glad to hear that!

> Do you want me to send a PR with point 1 and 2 above ? Or only point 1?

Let's got one by one, so the frist PR would be concerning the strategy name in the token.

> Do you agree with the way the strategy is stored in the code I propose ?

Yes, as far as I can tell now it's all fine. The only thing is that I would change the `strategyName` to 
`ownerStrategyName` so that it is clear that this parameter returns name of the strategy created this token instance.

alain-charles:
Ok Dmitry, gonna send 1st PR within a few hour.
Adding ownerStrategyName to NbAuthCreateToken is a breaking change.

Do you want me to make the parameter optionnal so that users car adapt their calling code in the next days ? 
I really think this parameter has to be mandatory in the long run .
What do you think ?

nnixaa:
I guess we can make it mandatory right away, as most of the time `NbAuthCreateToken` is used inside of the auth module and not directly by the user. For those who use it directly, we need to add a breaking change note to the commit message (https://github.com/akveo/nebular/blob/master/CONTRIBUTING.md#footer) so that it will be inserted into the changelog.md

alain-charles:
PR will be merged into multiples PR

PR #571, opened 2018-07-19
feature/auth : added a back link to the strategy name in the token (future use for instance refreshToken)

alain-charles:
The token now contains ownerStrategyName, which is a back link to the strategy
used to create the token (future use).

BREAKING CHANGE :
NbAuthCreateToken (token.ts) now takes a third parameter, which is the ownerStrategyName
Tokens are created by strategies that are called from services, so it is *potentially* a breaking change.


src/framework/auth/services/token/token.ts
```diff
@@ -17,11 +19,18 @@ export interface NbAuthRefreshableToken {
...
 export interface NbAuthTokenClass {
   NAME: string;
-  new (raw: any): NbAuthToken;
+  new (raw: any, ownerStrategyName: string): NbAuthToken;
 }
 
-export function nbAuthCreateToken(tokenClass: NbAuthTokenClass, token: any) {
-  return new tokenClass(token);
+// All types of token are not refreshables
+export function isNbAuthRefreshableToken(token: any): token is NbAuthRefreshableToken {
```
nnixaa: I guess this is not a part of this PR, isn't it?

src/framework/auth/services/token/token.ts
```diff
@@ -142,9 +156,11 @@ export class NbAuthOAuth2Token extends NbAuthSimpleToken {
 
   static NAME = 'nb:auth:oauth2:token';
 
-  constructor(protected data: { [key: string]: string|number }|string = {}) {
+  constructor(protected data: { [key: string]: string|number }|string = {},
+              protected ownerStrategyName: string) {
     // we may get it as string when retrieving from a storage
-    super(prepareOAuth2Token(data));
+    super(prepareOAuth2Token(data), ownerStrategyName);
+    this.ownerStrategyName = ownerStrategyName;
```
nnixaa: why to assign it here again as it will be assigned in `super`?

src/framework/auth/strategies/password/password-strategy.spec.ts
```diff
@@ -26,7 +26,7 @@ describe('password-auth-strategy', () => {
     },
   };
 
-  const successToken = nbAuthCreateToken(NbAuthSimpleToken, 'token');
+  const successToken = nbAuthCreateToken(NbAuthSimpleToken, 'token', 'strategy');
```
nnixaa: let's move name of the strategy `strategy` to a variable

src/framework/auth/services/token/token.ts
alain-charles: you're right. Part of the next next next :)

alain-charles:
@nnixaa it must be ok now.
Preparing the next PR for createdAt and expDate of the tokens.

Please let me know if you agree  with creating a NbAuthOAuth2JWTToken so that we have something symetric in token.ts and we can use in priority the payload 'iat' for createdAt and directly 'exp' for returning expDate.

best regards

Alain

nnixaa:
Hey @alain-charles, let's probably start with the `createdAt` property of a token, add it into the base token and then based on that create a new token?

Issue #564, opened 2018-07-15
Wrong implementation of "getTokenExpDate" in NbAuthOAuth2Token


schuettecarsten:
### Issue type

**I'm submitting a ...**  (check one with "x")

* [X] bug report
* [ ] feature request

### Issue description

**Related code:**
```
  /**
   * Returns expiration date
   * @returns Date
   */
  getTokenExpDate(): Date {
    if (!this.token.hasOwnProperty('expires_in')) {
      return null;
    }

    const date = new Date();
    date.setUTCSeconds(new Date().getUTCSeconds() + Number(this.token.expires_in));

    return date;
  }
```

Assuming token object will contain the number of seconds when the token expires in 'expires_in'. The number of seconds is valid at the moment when the token is received from the server. The client can use this information to calculate the exact expiration time of the token when the token is received.

The current getter implementation will add the number of seconds to the current date/time everytime the getter is called, which might be hours or days later and which will return an expiration date that's always in the future. This is wrong.

### Other information:

**Angular, Nebular**
```
Nebular 2.0.0-rc.9
```

alain-charles:
I agree the implementation above is wrong.
I sent a pull request three days ago concerning refreshToken management and i have proposed a correction for that because i needed it to be correct.
In the constuctor of the NbAuthOauth2Token, the proposed code search the 'exp' property in the token's payload if any, and if not found, uses the expires_in to calculate and store the expiration date.
```
    if (!this.expDate) {
      if (this.token && Object.keys(this.token).length > 0) {
        const decoded = this.getPayload();
        if (decoded.hasOwnProperty('exp')) {
          this.expDate = new Date(decoded.exp);
        }
      }
      if (!this.expDate) {
        if (this.token.expires_in) {
        this.expDate = new Date();
        this.expDate.setTime(this.expDate.getTime() + Number(this.token.expires_in) * 1000);
        }
      }
    }
```
If the code is merged by @nnixaa , i think this issue will be corrected.

schuettecarsten:
@alain-charles you mean your PR https://github.com/akveo/nebular/pull/560? It's currently blocked for merge :-(

alain-charles:
@schuettecarsten yes exactly.
The code has to be reviewed by nnixaa before he merges, especially because there are quite heavy changes in this PR and architectural questions.
So it may take a little time, but we are on our way.

chhatrapati:
Hi guys can you please help me in #566

schuettecarsten:
[...]

PR #576, opened 2018-07-24
Token now have createdAt attribute & issue #564 corrected

alain-charles:
#### feat/auth tokens now have the createdAt attribute:
- NbAuthSimpleToken = when backend sent it,
- NbAuthJWTToken : exp if exists, when backend sent it otherwise
- NbAuthOAuth2Token  = when backend sent it.

 #### resolved issues
Issue #564 

#### TODO:
- create a NbAuthOAuth2JWTToken (when Oauth2 backends send jwt token) ?
- create getAccessToken() method and reserve the getValue() method for the token as it was sent by the backend ?

### Remarks
Do not understand why so many files (11) changed when comparing to arkveo master ?


nnixaa:
Hey @alain-charles, could you please resolve the conflicts?

src/framework/auth/services/token/token-parceler.ts
```diff
@@ -6,6 +6,7 @@ import { NB_AUTH_TOKENS } from '../../auth.options';
 export interface NbTokenPack {
   name: string,
   ownerStrategyName: string,
+  createdAt: Date,
```
nnixaa: why Date if we set the time later?

src/framework/auth/services/token/token.ts
```diff
@@ -19,13 +20,14 @@ export interface NbAuthRefreshableToken {
 
 export interface NbAuthTokenClass {
   NAME: string;
-  new (raw: any, ownerStrategyName: string): NbAuthToken;
+  new (raw: any, ownerStrategyName: string, createdAt?: Date): NbAuthToken;
```
nnixaa: I believe createdAt shouldn't be optional here

src/framework/auth/services/token/token.ts
```diff
@@ -36,8 +38,26 @@ export class NbAuthSimpleToken extends NbAuthToken {
   static NAME = 'nb:auth:simple:token';
 
   constructor(protected readonly token: any,
-              protected readonly ownerStrategyName: string) {
+
+              protected readonly ownerStrategyName: string,
+              protected createdAt?: Date) {
     super();
+    if (!this.createdAt) {
```
nnixaa: probably we should go a bit different way?

```
this.createdAt = this.prepareCreatedAt(createdAt);
```
and then 

```
// this is for simple token
protected prepareCreatedAt(date: Date) {
  return date;
}

// this is for jwt
protected prepareCreatedAt(date: Date) {
  const decoded = this.getPayload();
  return decoded && decoded.ait ? new Date(Number(decoded.iat) * 1000) : date;
}
```

src/framework/auth/services/token/token.ts
```diff
@@ -119,10 +151,8 @@ export class NbAuthJWTToken extends NbAuthSimpleToken {
     if (!decoded.hasOwnProperty('exp')) {
       return null;
     }
-
     const date = new Date(0);
-    date.setUTCSeconds(decoded.exp);
-
+    date.setTime(Number(decoded.exp) * 1000); // 'cause jwt token are set in seconds
```
nnixaa: why setTime instead of setUTCSeconds?

src/framework/auth/services/token/token.ts
alain-charles: @nnixaa can you confirm what you mean : 
- `createdAt` is an optional parameter of the constructor
- if present, we use it to initialize the simpleToken `createdAt` attribute 
- if present, we use it to initialize the "without iat AuthToken " `createdAt` attribute 
- if not present, i think we have to use `new Date()` by default.

So i guess the code should be :
    `this.createdAt = this.prepareCreatedAt(createdAt);`

and then 
```
protected prepareCreatedAt(date: Date) {
    // For simple tokens, if not set the creation date is 'now'
    return date ? date : new Date();
  }
```
and for NbAuthJWTToken
```
  protected prepareCreatedAt(date: Date) {
    let decoded = null;
    try { // needed as getPayload() throws error and we want the token to be created in any case
      decoded = this.getPayload();
    }
    finally {
      return decoded && decoded.iat ? new Date(Number(decoded.iat) * 1000) : (date ? date : new Date());

    }
  }
```

What do you think ?

src/framework/auth/services/token/token.ts
alain-charles: Right

src/framework/auth/services/token/token-parceler.ts
alain-charles: You are right, reviews are helpful. 
I decided to store the date as time after i wrote this code.

src/framework/auth/services/token/token.ts
```diff
@@ -152,7 +181,9 @@ export class NbAuthOAuth2Token extends NbAuthSimpleToken {
   static NAME = 'nb:auth:oauth2:token';
 
   constructor(protected data: { [key: string]: string|number }|string = {},
-              protected ownerStrategyName: string) {
+              protected ownerStrategyName: string,
+              protected createdAd?: Date) {
```
nnixaa: createdAd -> createdAt

src/framework/auth/services/token/token.ts
```diff
@@ -80,6 +97,20 @@ export class NbAuthJWTToken extends NbAuthSimpleToken {
...
+   * for JWT token, the iat (issued at) field of the token payload contains the creation Date
+   */
+  protected prepareCreatedAt(date: Date) {
+    let decoded = null;
+    try { // needed as getPayload() throws error and we want the token to be created in any case
+      decoded = this.getPayload();
+    }
+    finally {
+      return decoded && decoded.iat ? new Date(Number(decoded.iat) * 1000) : (date ? date : new Date());
+
```
nnixaa: unnecessary new line
I would suggest the following change
 
```
protected prepareCreatedAt(date: Date) {
  date = super.prepareCreatedAt(date); // get or otherwise create
  let decoded = null;
  try { // needed as getPayload() throws error and we want the token to be created in any case
    decoded = this.getPayload();
  } finally {
    return decoded && decoded.iat ? new Date(Number(decoded.iat) * 1000) : date;
  }
}
```

src/framework/auth/services/token/token.ts
nnixaa: Yes this is correct, with the only code style change I left.

alain-charles:
@nnixaa 
NbAuthOAuth2Token constructor now looks like this 
```
constructor( data: { [key: string]: string|number }|string = {},
               ownerStrategyName: string,
               createdAt?: Date) {

    // we may get it as string when retrieving from a storage
    super(prepareOAuth2Token(data), ownerStrategyName, createdAt);
```
I removed the  `protected` keyword because these attributes are already declared in the ancestor, and on the other hand, this syntax was overriding the createdAt attirbute initialized in the ancestor (so that the createdAt attribute was undefined)

nnixaa:
@alain-charles, that makes sense. The only thing left is `prepareCreatedAt(` method, it still contains an unnecessary check `(date ? date : new Date());` which is moved into `super` call.

alain-charles:
Oh yes i'm so sorry ! I put it back for debugging the createdAt undefined explained above.
Coming back immediately !

alain-charles:
@nnixaa How do we go on now ?
The initial goal was to send automatic refresh tokens .

What do you think about these 2 points :

- create a NbAuthOAuth2JWTToken (case when Oauth2 backends send jwt tokens) ?

- create getAccessToken() method and reserve the getValue() method for the token as it was sent by the backend ?

nnixaa:
@alain-charles let's move to the custom ouath2jwt token. Regarding the second one - why do we need this change?

alain-charles:
OK let's go.
For the `getValue` and `getAccessToken` methods, i can't remember exactly, maybe we do not need both really.  We will see next days.

Issue #581, opened 2018-07-26
I gave the clientId and clientSecret but the request doesn't include Authorization header with Basic string 


rakesh-sharma:
I am using NbOAuth2AuthStrategy like below 

     
```
       name: 'password',
        clientId :'xxxx',
        clientSecret: 'xxxx',
        baseEndpoint: 'https://www.xxx.com/',
	token: {
            grantType: NbOAuth2GrantType.PASSWORD,
            class: NbAuthOAuth2Token,
            endpoint: 'oauth/token'
        }
```
It should ideally send the Authorization header with Basic string generated using clientId and clientSecret, but it's not sending is there any way I can send this header?

Also the payload has parameter with name 'email' instead of that Can I name it 'username'

alain-charles:
You are right, it is a TODO in the NbOAuth2AuthStrategy.

According to [https://tools.ietf.org/html/rfc6749#section-2.3](https://tools.ietf.org/html/rfc6749#section-2.3) :

- Clients in possession of a client password MAY use the HTTP Basic authentication scheme (...)
- Alternatively, the authorization server MAY support including the
   client credentials in the request-body using the following
   parameters: client_id (...) and client_secret (...)

As a contributor to this strategy, i'll propose  a PR that writes for the moment the first point, i.e. the basic authentication scheme.

@nnixaa what do you think ?

Alain

PR #582, opened 2018-07-26
feature/auth : NbOAuth2AuthStrategy implementing basic authentication scheme against token endpoints

alain-charles:
#### What it resolves
NbOAuth2Strategy now implements client authentication as specified in [RFC 6749 section 2-3](https://tools.ietf.org/html/rfc6749#section-2.3)

There is a new optional parameter of `NbOAuth2StrategyOption`.
The parameter is `clientAuthMethod`, and is a member of `NbOAuth2ClientAuthMethod ` enum:
- `NONE` (default) : no credentials are sent => No breaking change,
- `BASIC` : credentials are sent in the authorization header
- `REQUEST_BODY`: credentials are sent in the request body

AuthMethod is used (credentials are sent) when accessing to the authServer for :
- Getting token with `authorization_code` grant_type
- Getting token with `password` grant-type
- Getting token with `refresh_token` grant-type

RFC6749 says the client must not authenticate when hitting authorize endpoints, even if asking for a token. So nothing changed here, only clientId is sent in the url.
 
 #### issue resolved
issue #581


src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -282,6 +282,20 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
     return this.cleanParams(params);
   }
 
+  protected buildRequestsHttpOptions(): any {
```
nnixaa: let's make it optional?

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -4,7 +4,7 @@
  * Licensed under the MIT License. See License.txt in the project root for license information.
  */
 import { Inject, Injectable } from '@angular/core';
-import { HttpClient, HttpErrorResponse } from '@angular/common/http';
+import {HttpClient, HttpErrorResponse, HttpHeaders} from '@angular/common/http';
```
nnixaa: spaces

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
alain-charles: This makes sense since clientId and clientSecret are optional

nnixaa: Sorry, my comment was a bit misleading. What I meant is that probably we should add an option to the configuration as you proposed in the PR description indicating if we want to enable the base auth, pass clientid/cliensecret as body parameters or none of the above?

nnixaa: @alain-charles just checking if you saw my comment :)

alain-charles:
I have seen it in my mailbox

Going to setup the options none / auth / body 

Alain

src/framework/auth/strategies/oauth2/oauth2-strategy.spec.ts
```diff
@@ -9,9 +9,8 @@ import { HttpClientTestingModule, HttpTestingController } from '@angular/common/
 import { RouterTestingModule } from '@angular/router/testing';
 import { ActivatedRoute } from '@angular/router';
 import { NB_WINDOW } from '@nebular/theme';
-
 import { NbOAuth2AuthStrategy } from './oauth2-strategy';
-import { NbOAuth2GrantType, NbOAuth2ResponseType } from './oauth2-strategy.options';
+import {NbOAuth2ClientAuthMethod, NbOAuth2GrantType, NbOAuth2ResponseType} from './oauth2-strategy.options';
```
nnixaa: spaces :)

PR #583, opened 2018-07-27
feat/auth : created NbAuthOauth2JWTToken class for handling Oauth2 tokens encapsulating jwt access_token

alain-charles:
#### What it does
The class `NbAuthOauth2JWTToken` has been created for handling oAuth2 backends whose access-token are jwt-token

#### To do
Implements some methods on the oAuth2 refreshtoken when it is in jwt format too ?
(may be useful in the future when refreshToken expired : useless in this case to send request to the resource backend, redirect to login)


src/framework/auth/services/token/token.ts
```diff
@@ -251,3 +253,48 @@ export class NbAuthOAuth2Token extends NbAuthSimpleToken {
...
+ */
+export class NbAuthOAuth2JWTToken extends NbAuthOAuth2Token {
+
+  static NAME = 'nb:auth:oauth2:jwt:token';
+
+  /**
+   * Returns access token payload
+   * @returns any
+   */
+  getAccessTokenPayload(): any {
```
nnixaa: so the `getValue())` will return an encoded token and this method returns the decoded value, right?

so the complete token is encoded twice - the whole payload and the access token itself is also a jwt token, right?

alain-charles: As for every token class, the getValue() method returns the string you have to put in the authorization header (Bearer). So yes, in this case, it returns the encoded payload of the access token

I do not know if we can say "encoded twice" since the Oauth2 token has a payload that is not 'encoded'. But yes, the complete token has got a payload with an access-token, that is a jwt token, i.e. the access-token has got a encoded payload with 'iss', 'iat', 'exp'.

I'm sorry, so sorry,  but here i am not an IT head, so we have microsoft ADFS OAuth2 servers 😨 , and they put jwt access tokens in their Oauth2 token payload.


[See here](https://docs.microsoft.com/fr-fr/windows-server/identity/ad-fs/overview/ad-fs-scenarios-for-developers)

>Tokens used
>These scenarios make use of three token types:
> - id_token: A JWT token used to represent the identity of the user. The 'aud' or audience claim of the id_token matches the client ID of the native or server application.
> - access_token: A JWT token used in Oauth and OpenID connect scenarios and intended to be consumed by the resource. The 'aud' or audience claim of this token must match the identifier of the resource or Web API.
 >- refresh_token: This token is submitted in place of collecting user credentials to provide a single sign on experience. This token is both issued and consumed by AD FS, and is not readable by clients or resources.

nnixaa: but probably this part should be in the app code already, not library? So that the `getPayload` will return a complete token payload and then you have to deal with the other parts of the token on your side?

nnixaa: or otherwise we need to also have a `getIdTokenPayload()` method at least.

nnixaa: Well, just thought of it one more time, and probably this isn't that bad to put these methods inside of the JWTOAuth token, but then I propose the following:

```
getValue() - token string to put into request, access_token string in case of OAuth2
getRefreshToken() - refresh token string to put into request
getIdToken() - id token string to put into request
getAccessTokenPayload()
getIdTokenPayload()
getRefreshTokenPayload()
```

alain-charles: According to me, `access-token` is for the ressource server, and `id-token` is for the client app (maybe more ngx-admin than nebular, maybe...). So maybe here nothing for the framework.

However, if we decide *in the framework* that we use the `exp` and `iat` field of the jwt access-token to initialize the `createdAt` and `expDate` (future use : auth service sends automatically the refreshtoken request), so we need the implementation i propose. I agree we can in this case implements also the `id-token` side, so the client App will be able to obtain user information.

If we decide to consider that we have got an Oauth2 token *before all*, even if it encapsulates exp informations in its access-token, so we keep using `now` for `createdAt` and `expires_in` for computing `expDate`, and so i agree we do not need this implementation.

You've got a architecture choice to make :)

alain-charles: i think you helped me in my reflexion. If i had to make the choice, i would send this PR to the recycle bin 🤣

alain-charles: Sorry it seems we had reflexion at the same time...
So do you want me do do that ?
```
getValue() - token string to put into request, access_token string in case of OAuth2
getRefreshToken() - refresh token string to put into request
getIdToken() - id token string to put into request
getAccessTokenPayload()
getIdTokenPayload()
getRefreshTokenPayload()
```

nnixaa: Well, as we still have the default OAuth2 token implementation which is according to the spec, we can move forward with a complete implementation of OAuth2JWT token.

alain-charles: @nnixaa 
As far as i know, id_token JWT token is not part of the RFC  6749 oAuth2 spec. 
But Section 8.4 of RFC 6749 allows extensions.

=> So `id_token` is defined by an extension of Oauth2 called OpenId Connect (implemented by Microsoft ADFS 2016, Azure, Google...)

If we want to introduce `id_token`, we have to introduce a *new strategy* (that probably extends `NbOAuth2Strategy`, let us call it `NbOpenConnectIdStrategy`), with the new response_type  `id_token`, and any of the next combination : 
`id_token token`
`code id_token`
`code token`
`code id_token token`
`none `

[See the openid connect spec](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#id_token)
[See this excellent paper](https://medium.com/@darutk/diagrams-of-all-the-openid-connect-flows-6968e3990660)

We will then have to introduce a new token class for handling that, extending NbAuthOAuth2JWTToken, let's call it `NbConnectIdToken`.

It would be wonderful to succeed in doing that, but i think we have to make it in multiple PR (maybe a branch for working quiet ?). I would love to help writing that.

So maybe for the moment we can only merge the NbAuthOauth2JWT token as it is, fully compliant with rfc 6749, what do you think ?

nnixaa: @alain-charles yeah, that makes sense. Just a couple of related questions bellow.

src/framework/auth/services/token/token.ts
```diff
@@ -251,3 +253,48 @@ export class NbAuthOAuth2Token extends NbAuthSimpleToken {
...
+    return decodeJwtPayload(this.getValue())
+  }
+
+  /**
+   * for Oauth2 JWT token, the iat (issued at) field of the access_token payload
+   */
+  protected prepareCreatedAt(date: Date) {
+    let decoded;
+    try {
+       decoded = this.getAccessTokenPayload();
```
nnixaa: so in case when OAuth2 token is JWT, the `iat` fields is a part of JWT access token, right?

src/framework/auth/services/token/token.ts
```diff
@@ -251,3 +253,48 @@ export class NbAuthOAuth2Token extends NbAuthSimpleToken {
...
+  }
+
+  /**
+   * Returns expiration date :
+   * - exp if set,
+   * - super.getExpDate() otherwise
+   * @returns Date
+   */
+  getTokenExpDate(): Date {
+    if (this.getAccessTokenPayload() && this.getAccessTokenPayload().exp) {
```
nnixaa: and same here, `exp` is moved into the token payload?
if so, we need to make this `getAccessTokenPayload` call once and probably not cover it with a `null` check same way as for the other tokens

src/framework/auth/services/token/token.ts
```diff
@@ -101,13 +126,12 @@ export class NbAuthJWTToken extends NbAuthSimpleToken {
    * for JWT token, the iat (issued at) field of the token payload contains the creation Date
    */
   protected prepareCreatedAt(date: Date) {
-    date = super.prepareCreatedAt(date);
-    let decoded = null;
-    try { // needed as getPayload() throws error and we want the token to be created in any case
+    let decoded
```
nnixaa: ;

src/framework/auth/services/token/token.ts
alain-charles: Yes exactly. Redundant with expires_in of the 'containing' token

src/framework/auth/services/token/token.ts
alain-charles: Yes exp is present in the token.access-token payload
OK i'll see that

PR #593, opened 2018-07-31
Feat/auth improvement : refresh-token are now saved if not repeated by the backend refresh endpoint

alain-charles:
Feature/Auth : improvement on refreshable tokens
At log on, some backends send Oauth2 token with a refresh-token value.
But some of them do no repeat the refresh token value when you then hit the refresh token endpoint.
So up to now, we were losing the refresh-token so that we were unable to send more than once a refresh token request

This PR corrects this problem:  if necessary , it gets the existing refresh-token value in the current token and injects it in the refreshed token before storing it.

Hope it helps.


src/framework/auth/services/token/token.ts
```diff
@@ -16,11 +16,19 @@ export abstract class NbAuthToken {
 
 export interface NbAuthRefreshableToken {
   getRefreshToken(): string;
+  // Needed when refresh-token response does not repeat refresh_token value
```
nnixaa: @alain-charles so it is possible that after refreshing the token server may not return a new refresh token? Just looked through the RFC and this part looks vague.

In any case, I would propose to move this functionality into the `refreshToken()` method of the strategy. In this case, we have the current token and can extract the refresh part right there with no need for all additional checks.

What do you think?

alain-charles: I think that you are a amazing code architect :)

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -199,13 +199,17 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
     return this.http.post(url, this.buildRefreshRequestData(token), this.buildAuthHeader())
       .pipe(
         map((res) => {
+          const refreshedToken = <NbAuthOAuth2Token>this.createToken(res);
```
nnixaa: let's just move this into a new method:

```
createRefreshedToken(res): NbAuthRefreshableToken {
...
}
```
and then call it almost as before:
```
return new NbAuthResult(
            true,
            res,
            this.getOption('redirect.success'),
            [],
            this.getOption('defaultMessages'),
            this.createRefreshedToken(res));
```

alain-charles: @nnixaa 
I think you mean 
```
createRefreshedToken(res, existingToken): NbAuthOAuth2Token {
...
}
```
since we need the existing token for extracting the existing refresh_token

And NbAuthOAuth2Token instead of NbAuthRefreshableToken so that we return something assignable to NbAuthResult.token i.e NbAuthToken

nnixaa: @alain-charles yes, of course, you are right.
But in the second part - return `NbAuthOAuth2Token` - is this always true? as by design we can assign any token to any strategy, so it is not necessary `NbAuthOAuth2Token`, but rather `NbAuthToken`.

alain-charles: @nnixaa here is the best implementation i have found
```
  protected createRefreshedToken(res, existingToken: NbAuthRefreshableToken): NbAuthToken {
    const refreshedToken = this.createToken(res);
    if (isNbAuthRefreshableToken(refreshedToken)
        && !refreshedToken.getRefreshToken()
        && existingToken.getRefreshToken()) {
      refreshedToken.setRefreshToken(existingToken.getRefreshToken());
    }
    return refreshedToken;
  }
```

i think this is the one that is the most in the spirit of your architecture

src/framework/auth/strategies/oauth2/oauth2-strategy.ts
```diff
@@ -376,6 +377,16 @@ export class NbOAuth2AuthStrategy extends NbAuthStrategy {
     }, {}) : {};
   }
 
+  protected createRefreshedToken(res, existingToken: NbAuthRefreshableToken): NbAuthToken {
```
nnixaa: I believe we can leave this check as:

```
if (!refreshedToken.getRefreshToken() && existingToken.getRefreshToken()) {
```
as we can only have the `NbAuthRefreshableToken` token here, which presumes that both required get and set method exist because this method could only be called from the `refreshToken` one, which cannot be called for a non-refreshable token.
In case when the developer forces a call of `refreshToken` method for a non-refreshble token it will be a typescript level error, which is completely fine in this case.

alain-charles: `!refreshedToken.getRefreshToken()` does not compile as `NbAuthToken` (the return type) does not implement this method.
That is the reason why i added the `isNbAuthRefreshableToken(refreshedToken)` test

So what do you prefer : 
- should we return NbAuthOauth2Token  ?
- or just cast in the body of the method the refreshedToken to NbAuth0Auth2Token ?
- or leave the code like in #10d0b31 commit ?

nnixaa: @alain-charles sorry for the delay, wrote the comment last Thursday but forgot to submit.

alain-charles: @nnixaa with the intersection type you propose above, the problem remains on `refreshedToken`

```const refreshedToken = this.createToken(res);```

`this.createToken(res)` returns  `NbAuthToken` that does not implement `getRefreshToken()`.

Any other idea ? 

Alain

nnixaa: @alain-charles something like this probably? https://github.com/alain-charles/nebular/pull/1/files

nnixaa:
@alain-charles cool!

Issue #517, opened 2018-06-25
Invalid JWT token in local storage will crash App / NbAuthJWTToken 


schuettecarsten:
### Issue type

**I'm submitting a ...**  (check one with "x")

* [X] bug report
* [ ] feature request

### Issue description

**Current behavior:**
When login response of a NbPasswordAuthStrategy returns an broken JWT token, it is stored in the local storage without validation. On the next request, the app will fail because NbAuthJWTToken cannot parse the token.

**Expected behavior:**
A broken token should be rejected during login.

**Related code:**
See the exception that occurs when token from login response is invalid:
```
ERROR Error: The token yeah! is not valid JWT token and must consist of three parts.
    at NbAuthJWTToken.push../node_modules/@nebular/auth/services/token/token.js.NbAuthJWTToken.getPayload (token.js:112)
    at NbAuthJWTToken.push../node_modules/@nebular/auth/services/token/token.js.NbAuthJWTToken.getTokenExpDate (token.js:139)
    at NbAuthJWTToken.push../node_modules/@nebular/auth/services/token/token.js.NbAuthJWTToken.isValid (token.js:160)
    at MapSubscriber.project (auth.service.js:75)
    at MapSubscriber.push../node_modules/rxjs/_esm5/internal/operators/map.js.MapSubscriber._next (map.js:35)
    at MapSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.next (Subscriber.js:54)
    at RefCountSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber._next (Subscriber.js:77)
    at RefCountSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber.next (Subscriber.js:54)
    at Subject.push../node_modules/rxjs/_esm5/internal/Subject.js.Subject.next (Subject.js:47)
    at ConnectableSubscriber.push../node_modules/rxjs/_esm5/internal/Subscriber.js.Subscriber._next (Subscriber.js:77)
```

This exception is shown in Console log when the App tries to use the broken Token, the app crashes:
```
core.js:1601 ERROR Error: Uncaught (in promise): Error: The token yeah! is not valid JWT token and must consist of three parts.
Error: The token yeah! is not valid JWT token and must consist of three parts.
    at NbAuthJWTToken.push../node_modules/@nebular/auth/services/token/token.js.NbAuthJWTToken.getPayload (token.js:112)
[...]

PR #597, opened 2018-08-02
Feat/auth : Rejects malformed JWTToken and requireValidToken Option

alain-charles:
#### Short description

The framework now :
- rejects every malformed JWT tokens (i.e empty, not presenting 3 parts, not base64 encoded) whatever the value of `requireValidToken`,
- proposes a new `requireValidToken` at the endpoint level that is set to `false` by default,
- keeps on accepting to authenticate/register/refresh/token if the backend does not return any token, but only if  `requireValidToken` is set to `false` (rejects otherwise)
- rejects  every `!isValid()` token if `requireValidToken` is set to true,

#### important
`requireValidToken`  option is set to false by default so that there is no breaking changes. 
`failWhenNoToken`has been removed from password strategy as it was still not released and becoming redundant

#### issue resolved
Resolves the #517 issue


alain-charles:
@nnixaa the last commit resolves merging conflict and implements failWhenNoToken removal.

src/framework/auth/services/token/token.ts
```diff
@@ -14,6 +14,14 @@ export abstract class NbAuthToken {
   }
 }
 
+export class InvalidJWTTokenError extends Error {
+  constructor(message: string) {
+    super(message);
+    const actualPrototype = new.target.prototype;
```
nnixaa: no need in additional variable here, just `Object.setPrototypeOf(this, new.target.prototype);`

src/framework/auth/services/token/token.ts
```diff
@@ -14,6 +14,14 @@ export abstract class NbAuthToken {
   }
 }
 
+export class InvalidJWTTokenError extends Error {
```
nnixaa: `InvalidJWTTokenError` -> `NbAuthInvalidJWTTokenError` would be better

src/framework/auth/services/token/token.ts
```diff
@@ -14,6 +14,14 @@ export abstract class NbAuthToken {
...
+  constructor(message: string) {
+    super(message);
+    const actualPrototype = new.target.prototype;
+    Object.setPrototypeOf(this, actualPrototype);
+  }
+}
+
```
nnixaa: we also need another error then, as we cannot use `InvalidJWTTokenError` everywhere, since not every token is JWT token.
probably like this
`NbAuthInvalidTokenError`
`NbAuthInvalidJWTTokenError`

src/framework/auth/services/token/token.ts
```diff
@@ -221,7 +237,7 @@ export class NbAuthOAuth2Token extends NbAuthSimpleToken {
    */
   getPayload(): any {
     if (!this.token || !Object.keys(this.token).length) {
-      throw new Error('Cannot extract payload from an empty token.');
+      throw new InvalidJWTTokenError('Cannot extract payload from an empty token.');
```
nnixaa: this isn't a JWT token error I suppose as we are in `NbAuthOAuth2Token`

src/framework/auth/strategies/auth-strategy.ts
```diff
@@ -21,7 +21,12 @@ export abstract class NbAuthStrategy {
   }
 
   createToken<T extends NbAuthToken>(value: any): T {
-    return nbAuthCreateToken<T>(this.getOption('token.class'), value, this.getName());
+    const requireValidToken: boolean = this.getOption('requireValidToken');
+    const token = nbAuthCreateToken<T>(this.getOption('token.class'), value, this.getName());
+    if (requireValidToken && !token.isValid()) {
+      throw new InvalidJWTTokenError('Token is empty or invalid.');
```
nnixaa: same here, we can only throw a more generic error here

src/framework/auth/strategies/dummy/dummy-strategy-options.ts
```diff
@@ -7,7 +7,6 @@ import { NbAuthStrategyOptions } from '../auth-strategy-options';
 import { NbAuthSimpleToken } from '../../services/';
 
 export class NbDummyAuthStrategyOptions extends NbAuthStrategyOptions {
-  name: string;
```
nnixaa: why do we need to remove this name?

src/playground/auth/auth.module.ts
```diff
@@ -74,7 +74,7 @@ import { NbAuthGuard } from './auth-guard.service';
 
         NbPasswordAuthStrategy.setup({
           name: 'email',
-
+          requireValidToken: false,
```
nnixaa: I would say we need to introduce this not on a strategy level, but on an endpoint level.
For instance, in a case when login endpoint requires to have a valid token, but register endpoint doesn't.

src/playground/oauth2-password/oauth2-password.module.ts
```diff
@@ -16,13 +16,11 @@ import {
 } from '@nebular/theme';
 
 import {
-  NbAuthModule,
+  NbAuthModule, NbAuthOAuth2JWTToken,
```
nnixaa: I guess this change should not be a part of this PR?

src/framework/auth/strategies/dummy/dummy-strategy-options.ts
alain-charles: @nnixaa 
As for the moment `requireValidToken` was set at strategy level, i put it in the ancestor `NbAuthStrategyOption`.
I then realized that only `NbPasswordStrategyOption` was extending `NbAuthStrategyOption`.
I guess it would be better (even if `requireValidToken` finally goes at endpoint level) to make OAuth2 and Dummy options also extend `NbAuthStrategyOption`.

In this case `name` (and probably more attributes) can be declared in `NbAuthStrategyOption`.

What do you think ?

alain-charles:
@nnixaa i'm still working on this.
Issue #517 i said 

> If the jwt token is malformed, i don't think we shoud accept it and store it, because it is really useless. we won't be able to use it for authenticating against backend.
> 
> If the token is malformed, maybe it is a misconfiguration of the strategy, maybe a backend problem.
> I propose that instead of crashing, we send NbAuthResult with success=false, and the error description "invalid jwt token".

Do you still agree with that, i.e.  even if requireValidToken is set to false on the endpoint, we reject the token if it is malformed ?

I still think we shoud reject malformed JWT tokens, whatever the value of `requireValidToken`. For me `requireValidToken` must be used to stop the process when we got no token at all or when it is expired.

What do you think ?

src/framework/auth/strategies/dummy/dummy-strategy-options.ts
nnixaa: @alain-charles I'm bit confused, as far as I can see, all strategies' options (oauth2, dummy, password) are extended from `NbAuthStrategyOptions`. Am I missing something?

nnixaa:
@alain-charles yes, let's do this way. But then we need to parse the token at the creation time I guess, not at the `getPayload` time?

src/framework/auth/strategies/dummy/dummy-strategy-options.ts
alain-charles: @nnixaa  Have a look at `NbOAuth2AuthStrategyOptions` in the master branch

alain-charles:
@nnixaa it is already the case in what i've written and i'll push
JWT token payload is parsed at creation time, the parsing is launched by prepareCreatetAt called in every constructor.
=> We have to fail if token id returned by backend but is invalid, 
=> and we have to create the token if no token is given because of requiValidToken than can be false

nnixaa:
@alain-charles yes, I agree, but since we parse it here already, moreover we have to parse it (to validate and fail) there is no need to parse it all over again on each `getPayload()` call. It would be better to parse/validate it in the constructor once and then use the parsed value in the consequential calls. What do you think?

alain-charles:
@nnixaa concretely you wanna store the decoded payload as a token attribute ?

nnixaa:
@alain-charles yes, something like this:

```
   constructor(protected readonly token: any,
               protected readonly ownerStrategyName: string,
               protected createdAt?: Date) {
     super();
	 this.payload = this.parsePayload();
     this.createdAt = this.prepareCreatedAt(createdAt);
   }

```

and then the `getPayload()` method is just a getter:

```
   getPayload(): string {
     return this.payload;
   }
```

Does it make sense?

alain-charles:
@nnixaa finally i suuceeded in pushing something that makes sense.
It was quite hard to have something working well in every case, but i think now we are (almost?) good.

nnixaa:
@alain-charles awesome, checking

src/framework/auth/strategies/password/password-strategy.ts
```diff
@@ -339,84 +292,63 @@ export class NbPasswordAuthStrategy extends NbAuthStrategy {
...
-
-        throw new Error('Could not extract token from the response.');
-      }
-      return res;
-    });
+  protected handleResponseError(res: any, module: string): Observable<NbAuthResult> {
+    let errors = [];
+    if (res instanceof HttpErrorResponse) {
+      errors = this.getOption('errors.getter')(module, res, this.options);
+    } else if (res instanceof NbAuthIllegalTokenError) {
```
nnixaa: so in case when the token is empty any of our token classes will throw the `NbAuthTokenNotFoundError` exception, right? Which won't be caught by this `if` and will go to `something went wrong` part?

In this case the statement `It MAY be created even if backend did not return any token, in this case, it is !Valid` is not valid and `empty` token will never be created.
If my reasoning is correct, then we just need to decide, what `empty` means - malformed or invalid.

alain-charles: @nnixaa here is what i have coded (or what i think i've coded 🤣 ) : 

- `NbAuthTokenNotFoundError` is thrown if we did not find the token key in the response payload (i.e. no `access_token` key in OAuth2 token response, no `token` key in simple token response...). That is what i call an `empty` token. 
=> the error is not catched during token construction, because we want to allow some endpoints to create such tokens if `requireValidToken` is false.
=> But after token construction, if `requireValidToken` is set to true, then the isValid() will return false so that the `createToken()` method will throw an `NbInvalidTokenError` saying the token is empty or not valid. `NbInvalidTokenError` will then be catched by the strategy that will display the error message.
```
if (failWhenInvalidToken && !token.isValid()) {
      // If we require a valid token (i.e. isValid), then we MUST throw NbAuthIllegalTokenError so that the strategies
      // intercept it
      throw new NbAuthIllegalTokenError('Token is empty or invalid.');
    }
```

- If we find something in the token key in the response payload (i.e `access_token` is present, `token` is present), then i consider that 
1/ we have a token (even if it is `''` or `{}`) 
2/ the token we have must respect some rules (JWT format for instance) and if not we throw NbIllegalTokenError that will be catched in every case by the strategies.


To conclude 
- `empty` means the key was not found in the response. 
- And `empty`tokens are never valid (`isValid()`) but can be created if `requireValidToken` is set to false
- If the key is found, the token is not `empty` even if its value is`''`or `{}` and then it must pass the "wellFormed" control otherwise it is declared Illegal (`NbIllegalTokenError`) and rejected.

Hope i am clear.

alain-charles: I think we can decide that `''` or `{}` are empty tokens....

nnixaa: Ah okay, so we catch the empty token error during this check, right?
```
    try {
      this.parsePayload();
    } catch (err) {
      if (err instanceof NbAuthIllegalTokenError) { // token is present but illegal (empty or malformed), we reject it
        throw err;
      }
    }
```
so if it is illegal - throw next, and if it's just invalid (empty) - do nothing, right?

alain-charles: Exactly!
If empty the constructor does not throw any error .
Then the createToken() method (that launched the constructor) :
- do nothing if requireValidToken is false so that we have an empty token created, stored and the login/register/refresh succeeds
- throw NbIllegalTokenError (we could create NbRequireValidTokenError) so that stratégies can return the message in NBAuthResult and login/refresh/register fail with the message

nnixaa: Got it, this is cool, but I'd suggest we inverse this check so that we don't block any other possible exception. So we specifically block the `NbAuthTokenNotFoundError` and don't touche other expectations including `NbAuthIllegalTokenError`:
```

 try {
      this.parsePayload();
    } catch (err) {
      if (!(err instanceof NbAuthTokenNotFoundError)) {
        throw err;
      }
    }

```

Or, another option is to not throw the `NbAuthTokenNotFoundError` expectation at all, just create the token.

alain-charles: Yes I agree .
I’ m gonna push very soon

alain-charles: @nnixaa What do you think now and what else can we do to be ready ?

PR #638, opened 2018-08-15
Feat/Auth : refreshToken requests fully managed by NbAuthService & NbAuthJWTInterceptor

alain-charles:
### RefreshToken requests fully managed by NbAuthService & NbAuthJWTInterceptor

This pull requests intends to provide the necessary code so that refreshtoken requests are automatically sent to the backend when the local stored accesstoken has expired.
This allows the user not to type his login/password as long as his refresh-token is valid.
The refreshtoken requests are sent transparently.

### Run it in the playground
You can easily test the code by calling the url http://localhost:4200/#/auth/api-calls.component
Then the AuthGuard service redirects you to the login page.

The strategy used to log in is Oauth2 with password grant-type, so you can use example@akveo.com.

You are then logged with an access token whose ttl is 60 seconds and a refresh token whose ttl is 120 seconds.
These values can be modified in the configuration of the backend (config.js).

capture d ecran 2018-08-15 a 10 04 43

### Call of the secured api
Once logged, go immediately to http://localhost:4200/#/auth/api-calls.component and click the call API button.

The XHR request is send with the authorization header containing Bearer + the access-token.
You can see that Alain's got good wines in his cellar.

capture d ecran 2018-08-15 a 10 05 40

### Automatic refresh-token request
Wait a minute, and click once again on the call-api button.

Observe the network logs of you browser : the acces-token was expired, so the framework sent automatically a refresh request to the backend
capture d ecran 2018-08-15 a 10 06 25
with the next content :
capture d ecran 2018-08-15 a 10 07 02
and it got a new access-token:
capture d ecran 2018-08-15 a 10 07 10

then it called with success then api.

### Automatic redirect to login
If you wait once a minute, and if you click the call-api button, the framework sends a new refreshToken request
capture d ecran 2018-08-15 a 10 07 41

 but you are redirected to the login page, because refresh-token itself expired.
capture d ecran 2018-08-15 a 10 07 56

Hope you will like this work.


src/framework/auth/services/auth.service.ts
```diff
@@ -21,8 +21,21 @@ import { NbAuthToken } from './token/token';
...
 export class NbAuthService {
 
+  private authEndPoints: string[] = ['login', 'register', 'request-pass', 'logout', 'token', 'refresh'];
+
+  // Authorization server urls - a way to make them available in http_interceptor or anywhere else
+  authUrls: string[] = [];
+
   constructor(protected tokenService: NbTokenService,
               @Inject(NB_AUTH_STRATEGIES) protected strategies) {
+    for (const strategy of this.strategies) {
```
nnixaa: To be honest I don't like this idea of having this kind of state in the auth service. I would say that the public urls should not be configured by the strategy endpoints, but rather separately inside of the interceptor. This way we make it explicit.

src/framework/auth/services/interceptors/jwt-interceptor.ts
```diff
@@ -13,24 +14,41 @@ export class NbAuthJWTInterceptor implements HttpInterceptor {
   }
```
nnixaa: so we can pass the public urls here as an injectable token for this particular interceptor.

src/framework/auth/services/auth.service.ts
```diff
@@ -34,12 +47,29 @@ export class NbAuthService {
   }
 
   /**
-   * Returns true if auth token is presented in the token storage
+   * Returns true if valid auth token is present in the token storage.
+   * If not, calls the strategy refreshToken, and returns true if it get a new valid access token, false otherwise
    * @returns {Observable<any>}
    */
   isAuthenticated(): Observable<boolean> {
```
nnixaa: I guess this first of all should be under a parameter `refresh = false` or even a separate method, like `isAuthenticatedOrRefresh()`.

src/framework/auth/services/auth.service.ts
```diff
@@ -34,12 +47,29 @@ export class NbAuthService {
...
     return this.getToken()
-      .pipe(map((token: NbAuthToken) => token.isValid()));
+      .pipe(
+        switchMap(token => {
+        if (!token.isValid() && (token.getOwnerStrategyName())) {
+          return this.refreshToken(token.getOwnerStrategyName(), token)
+            .pipe(
+              switchMap(res => {
+                if (res.isSuccess()) {
+                  return observableOf(res.getToken().isValid());
```
nnixaa: and then if isSuccess === true, we should call the isAuthenticated method again without refreshing a token, just to make sure that the whole process behind went well.

src/framework/auth/services/auth.service.ts
alain-charles: @nnixaa yes i agree with you.
I had a long time asking myself how to make the auth urls available in the interceptor so that it does not intercept them.
I choosed not to ask the user to declare twice the auth urls, once in the strategy descripton, once elsewhere so that it could be injected or available in the interceptor.

So you agree with the fact that nebular users declare twice the auth urls ?

nnixaa: I guess we shouldn't call these `auth urls`, these are rather `public urls`, which means that they are not limited to the urls used for the auth. Thus there should be a way to provide this list anyway.

src/framework/auth/services/auth.service.ts
alain-charles: Yes it will be clearer .

src/framework/auth/services/auth.service.ts
alain-charles: I think I got it.
You are saying that in every case the user has to provide a list of public urls so that it is not very embarrassing to ask him to add auth urls to his list ?

alain-charles:
@nnixaa and akveo team,

It is more complicated than it seemed to be. Here is my point of view of the thing.

The framework we are building actually knows things about the token it obtained after auth. 
More precisely, it seems it was decided to implement the ability to refresh the token, since we find the refreshToken method in every strategy implementing this kind of re-auth.
NbAuthService itself implements refreshToken. So we could say that it is written for the client code only, but i am not sure. Let's take the above arguments :

> Token interceptor should know nothing about public URL in most of the cases (except for some rare cases). 

I completely agree with that. That is the reason why i first proposed to make available in the interceptor the `auth urls` (for infinite loop troubleshooting), not the `public urls`.

Consider now the second point, we have your 3 cases :

> - we have a valid token, we do a request, the token is valid on backend, OK
> - we have an invalid token, we don't do a request, we redirect to the login, OK
> - we have a valid token, we do a request, but on the backend the token was just rejected 

I still agree with that, but if we want to be honnest, we have to add 2 more cases : 
- We have an invalid token, we know we have a refresh token, we know the url it can be posted  to obtain a valid token...and we should do nothing ? => that is not what i think, i think we should refresh then send the request with the new access token and this implies the `interceptor` or anyone else knows the `auth-urls` if we do not want to fall into infinite loop
- We have an invalid token, we know we have a refresh token, we know the url it can be posted  to obtain a valid token, we post it, and we get a invalid refresh token error...=> the response must be treated, same as you third case.

What do yout think if 
- We find a way to make `auth-url` available to interceptor because we can't send refresh token requests that would be themselves intercepted and so on to infinite loop,
- We call NbAuthService.isAuthenticatedOrRefresh() 
  - true : we send the request `next.handle()` and we catchError
     - OK
     - catchError with `error instanceOf HttpErrorResponse` and `error.status === 401` => Redirect to login
  - false => Redirect to login

In this case, the NbAuthService is used to refresh the token if necessary (and i think this is really its work), and the interceptor puts the token in the header, catch 401/403 errors ans redirects in this case to login.

Does it seem acceptable to you ?

nnixaa:
@alain-charles right, that's the point we were missing!

Okay, I don't have a proper solution now... But really I don't like the idea of `auth urls` as then the auth service knows a lot more then it should do. Moreover, we tightly couple the auth service and interceptor, which we shouldn't do.

The best solution I have in mind now - some kind of `exclude urls` for the interceptor. But I'm not sure about that, let's sleep on it till tomorrow :)

But the redirect thing - I'd suggest we move it from this interceptor and the framework at all. Or I'm missing something why we should keep it inside?

alain-charles:
@nnixaa & team

> But really I don't like the idea of auth urls as then the auth service knows a lot more then it should do. Moreover, we tightly couple the auth service and interceptor, which we shouldn't do.

I completely agree but up to now i haven't found a better way to do it without asking the user to repeat the auth urls.  But i am ready to setup a solution when the user has to precise which url should be excluded from interceptor, even if he has to repeat things twice, cause nebular must have the best possible architecture .

> But the redirect thing - I'd suggest we move it from this interceptor and the framework at all. Or I'm missing something why we should keep it inside?

You are, once again, right, and the discussion we have is rich of all these things we say to each other.
So you convinced me at 100% that in case of 401/403 coming from `next.handle()`, or if `isAuthenticatedOrRefresh()` returns `false`, the interceptor should FAIL (i.e. throw the ErrorResponse) so that the CLIENT code do what it wants with that (and probably redirect to login).

I think we are fine now. The framework (`NbAuthService`) does everything it can with its token knowledge to maintain the authentication, puts the token in the requests header, and fails if it cannot, letting the client code doing its stuff.

It sounds really good to me now. We only have to decide now how to avoid the infinite loop in the interceptor when it calls auth urls . 
I hope the night will help us.

alain-charles:
@nnixaa I propose we create an Injectable NbAuthUrls class that stores the strategies base endpoints.
The only instance (singleton) is created in auth.options.ts
```
export const NB_AUTHURLS = new NbAuthUrls();
```
Then the client code provides it as it is the case with NbAuthJWTInterceptor in its client auth.module
```
    { provide: HTTP_INTERCEPTORS, useClass: NbAuthJWTInterceptor, multi: true },
    { provide: NbAuthUrls, useValue: NB_AUTHURLS},
```

Whe have to feed it in the strategy factory that already knows the Injector
```
export function nbStrategiesFactory(options: NbAuthOptions, injector: Injector): NbAuthStrategy[] {
  const strategies = [];
  const authUrls = injector.get(NbAuthUrls);
  options.strategies
    .forEach(([strategyClass, strategyOptions]: [NbAuthStrategyClass, NbAuthStrategyOptions]) => {
      const authBaseUrl = strategy.getOption('baseEndpoint');
      if (authBaseUrl && !authUrls.getUrls().includes(authBaseUrl)) {
        authUrls.add(authBaseUrl);
      }
      strategies.push(strategy);
    });
  return strategies;
}
```

then we inject it in the NbAuthJWTInterceptor.

The user has only to add the line in its client code, and does not repeat the authUrls.

Is it acceptable ?

alain-charles:
It would be perfect if the list of items of the filter function could be initialized by client code and completed by the framework with authurls baseEndpoints.

According to you, where  should the client code add its own urls that must not be injected a token ?
How could we do that with a simple function injected and not with a class permitting url adds ?

Shall we calll the NbAuthUrls class NbAuthInterceptorFilter ans add a filter method called by interceptor ?

alain-charles:
Maybe a bit heavy but why not setting  up a NbAuthJwtInterceptorOptions as de do for strategies and the options could be 
- filterAuthUrls: boolean 
- urlsToFilter: string[] 

??

alain-charles:
I think we are falling back to ask the interceptor donti what is it not made for : shouldn’t we only filter authurls as we said yesterday ?

nnixaa:
@alain-charles sorry I'm probably missing the point of:
```
According to you, where should the client code add its own urls that must not be injected a token ?
How could we do that with a simple function injected and not with a class permitting url adds ?
```

As auth urls are configured by the user, that means that he/she can simply provide this check on this urls in a function. So yes, that means the developer must duplicate this urls in both configurations, but I don't think this is very bad since this is all configured in one place, so even could be moved in some variable but the developer, for example into the environment variable.

alain-charles:
@nnixaa i am sorry i'm not clear enough and my english is far from being perfect.

So if i understand well, you propose that : 
- in the *client* `auth.module.ts` (ie where the user setup his/her strategies and so on), the user also writes the filter function and injects it,
- the `NbAuthJwtTokenIngterceptor` calls this injected filter function.

If so, i  we will probably encounter following troubles : 
- if the user omits to provide the filter function, the interceptor will not work (trying to get a token that was not injected) => we can decide it is the user's problem and document it clearly,
- if the user provides the filter function but omits to include auth urls, the interceptor will go into infinite loop. Do you think it is acceptable ?

But maybe i did not undestand exactly what you meant.

alain-charles:
@nnixaa i have here on my computer two implementations :
- first one tight on `auth urls` with `NbAuthUrls` class initialized in `strategy factory` and injected in interceptor.
- second one implementing and injecting the filter function in the `NbAuthPlaygroundModule`

Both works perfectly.
- The first one does not need the user to add anything in his configuration but any other urls are injected a token in their authorization header,
- the second one needs the user to write his filter function, without omitting the auth urls, but is more flexible as the user can add his own urls.

To go on, please let me know which one you want me to push ?

nnixaa:
@alain-charles yes, I'm okay with it. The code indeed requires some cleanup but I liked it as a PoC.

Also, I would also divide this PR in two (sorry I know that might be frustrating that we cannot complete it for that long, but I believe it worth it) - the actual changes + test, and the playground example (so we have a better commits history in GIT).

alain-charles:
Ok i’ll clean the code and send a first PR tomorrow morning .
Then when merged, i will push a second PR with playground example .

nnixaa:
@alain-charles closing this one

BlackXnt:
Is it possible to set up with NbPasswordAuthStrategy?

PR #649, opened 2018-08-18
Feat/Auth : refreshToken request automatically sent by NbAuthService if isAuthenticatedOrRefresh is called

alain-charles:
Feat/Auth :

- refreshToken request automatically sent by `NbAuthService` if `isAuthenticatedOrRefresh()` is called
- `NbAuthJWTInterceptor` now calls `NbAuthService.isAuthenticatedOrRefresh()`

POTENTIAL BREAKING CHANGE if `NbAuthJWTInterceptor` is used :
The developper has now to provide in his auth_module a function `filterInterceptorRequest(req: HttpRequest<any>):boolean` that returns true if req.url must not be intercepted for bearer token injection.
The auth urls MUST be included there.

```
export function filterInterceptorRequest(req: HttpRequest<any>) {
  return ['http://localhost:4400/api/auth/',
              'http://other.url/with/no/token/injected/',
         ]
    .some(url => req.url.includes(url));
}
....
{ provide: NB_AUTH_TOKEN_INTERCEPTOR_FILTER, useValue: filterInterceptorRequest},
```

Playground sample follows.


src/framework/auth/services/auth.service.ts
```diff
@@ -34,14 +34,40 @@ export class NbAuthService {
...
+                if (res.isSuccess()) {
+                  return this.isAuthenticated();
+                } else {
+                  return observableOf(false);
+                }
+              }),
+            )
+        } else {
+          return observableOf(token.isValid());
+          }
```
nnixaa: just an unnecessary space

src/framework/auth/services/auth.spec.ts
```diff
@@ -24,7 +24,7 @@ describe('auth-service', () => {
   let tokenService: NbTokenService;
   let dummyAuthStrategy: NbDummyAuthStrategy;
   const testTokenValue = 'test-token';
-  const ownerStrategyName = 'strategy';
+  const ownerStrategyName = 'dummy';
```
nnixaa: do we need this change? looks unrelated

src/framework/auth/auth.module.ts
```diff
@@ -60,7 +60,6 @@ export function nbStrategiesFactory(options: NbAuthOptions, injector: Injector):
     .forEach(([strategyClass, strategyOptions]: [NbAuthStrategyClass, NbAuthStrategyOptions]) => {
       const strategy: NbAuthStrategy = injector.get(strategyClass);
       strategy.setOptions(strategyOptions);
-
```
nnixaa: do we need this change? looks unrelated

src/framework/auth/services/interceptors/jwt-interceptor.ts
```diff
@@ -2,9 +2,9 @@ import { Injectable, Injector } from '@angular/core';
 import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest } from '@angular/common/http';
 import { Observable } from 'rxjs';
 import { switchMap } from 'rxjs/operators';
-
+import { NbAuthToken} from '../token/token';
 import { NbAuthService } from '../auth.service';
-import { NbAuthJWTToken } from '../token/token';
+import { NB_AUTH_TOKEN_INTERCEPTOR_FILTER} from '../../auth.options';
```
nnixaa: space

src/framework/auth/services/interceptors/jwt-interceptor.ts
```diff
@@ -13,24 +13,44 @@ export class NbAuthJWTInterceptor implements HttpInterceptor {
...
-        }),
-      );
+    // do not intercept request whose urls are filtered by the injected filter
+      if (!this.filter(req)) {
+        return this.authService.isAuthenticatedOrRefresh()
+          .pipe(
+            switchMap(authenticated => {
+              if (authenticated) {
+                  return this.authService.getToken().pipe(
+                    switchMap( (token: NbAuthToken) => {
```
nnixaa: space

src/framework/auth/services/interceptors/jwt-interceptor.ts
```diff
@@ -13,24 +13,44 @@ export class NbAuthJWTInterceptor implements HttpInterceptor {
...
+      return next.handle(req);
+    }
   }
 
   protected get authService(): NbAuthService {
     return this.injector.get(NbAuthService);
   }
+
+  protected get filter(): (req: HttpRequest<any>) => boolean {
+    return this.injector.get(NB_AUTH_TOKEN_INTERCEPTOR_FILTER);
```
nnixaa: I guess the token can be injected using constructor (except for AuthService)

src/framework/auth/services/interceptors/jwt-interceptor.ts
```diff
@@ -13,24 +13,44 @@ export class NbAuthJWTInterceptor implements HttpInterceptor {
...
+  }
+
 }
+
+
+
```
nnixaa: couple of unnecessary new lines

src/framework/auth/services/auth.spec.ts
alain-charles: Yes we do since i added `isAuthenticatedOrRefresh()` calls in the test units.
The strategy provided in test module is named 'dummy'.
So i modified the ownerStrategyName to `dummy` and replaced all 'dummy' occurrences by `ownerStrategyName`.
I know everything is not related to this PR, but is was necessary.

src/framework/auth/auth.module.ts
alain-charles: OK i put on back the empty line :)

src/framework/auth/services/interceptors/jwt-interceptor.ts
nnixaa: @alain-charles just one last formatting issue here :)
`switchMap( (token: NbAuthToken) => {` -> `switchMap((token: NbAuthToken) => {`

alain-charles: I run npm lint before I push but it seems my config is not as restrictive as yours .
Which one do you use ?

nnixaa: It looks like the project lint config isn't the best one. Will look into it.

nnixaa:
@alain-charles I just noticed that we don't provide a default value for the filter right? Meaning that if we try to use the interceptor without providing a filter it will fail by calling `this.filter()` right?

alain-charles:
No we don’t.
That is he potential breaking change I am talking about in the PR Desc.
I agree we should provide a no op filter.
I’ll push very soon

src/framework/auth/services/auth.service.ts
```diff
@@ -34,14 +34,40 @@ export class NbAuthService {
...
+   * Returns true if valid auth token is present in the token storage.
+   * If not, calls the strategy refreshToken, and returns isAuthenticated() if success, false otherwise
+   * @returns {Observable<boolean>}
+   */
+  isAuthenticatedOrRefresh(): Observable<boolean> {
+    return this.getToken()
+      .pipe(
+        switchMap(token => {
+        if (!token.isValid()) {
+          return this.refreshToken(token.getOwnerStrategyName(), token)
```
cloakedch: Does this work with any kind of token type? Is the refreshToken endpoint always expecting the same token that was stored?
I am using an API that during authentication returns an object consisting of a refresh token, an access token and other properties. However, on refreshing the token I am expected to only supply the refresh token.

To get it working, I had to convert the token to an NbAuthOAuth2Token and explicitly call "oAuth2TOken.getRefreshToken()" on this line instead of just "token".

alain-charles: Hi @cloakedch ,

The refresh_token is implemented according to [RFC6749 section 1.5](https://tools.ietf.org/html/rfc6749#section-1.5).

This RFC specifies the OAuth2 authentication scheme. So the code works with auths server returning tokens according to the [RFC section 4.2](https://tools.ietf.org/html/rfc6749#section-4.2.2)
The refreshToken endpoint is waiting for the refresh_token that was sent during login or during the latest refresh_token call. Both `grant_type=refresh_token` and `refresh_token=theRefreshToken`  are sent to the refreshToken endpoint for getting a new access-token, according to [the RFC](https://tools.ietf.org/html/rfc6749#page-47).

I would say that you  have to use the `NbOAuth2AuthStrategy`  with `password` grantType(if you send login/password at login) and `NbAuthOAuth2Token`class in your case.

Someting like
```
            NbOAuth2AuthStrategy.setup({
          name: 'password',
          clientId: 'test',
          clientSecret: 'secret',
          baseEndpoint: 'http://your.BaseEnd/Point',
          token: {
            endpoint: 'token',
            grantType: NbOAuth2GrantType.PASSWORD,
            class: NbAuthOAuth2Token,
          },
          refresh: {
            endpoint: 'refresh-token',
            grantType: NbOAuth2GrantType.REFRESH_TOKEN,
          },
        }),
```
Hope it helps

cloakedch: OK, I guess the API I am using is then not working according to the RFC.
Thanks for clearing that up!

nnixaa:
@alain-charles congrats!

Issue #540, opened 2018-06-29
Tabs are shifted on hover


EatonZ:
### Issue type

**I'm submitting a ...**  (check one with "x")

* [x] bug report
* [ ] feature request

### Issue description

**Current behavior:**
When hovering over tabs, other tabs to the left or right will shift, which is very distracting.

**Expected behavior:**
Hovering over a tab should not affect the other tabs at all.

**Steps to reproduce:**
This can be seen on the [demo page](https://akveo.github.io/nebular/docs/components/tabs/overview). In the first example, hover over the Orders tab and you'll see the Transactions tab shift a little to the right. In the third example, hover over Orders and you will see Users change slightly.

### Other information:

**Browser:** Chrome 67 on Windows 10 1803

PR #651, opened 2018-08-18
fix(tabset): hovering a tab doesn't affect the other tabs (#540)

cmajsmith:

This is a fix for #540 issue.


nnixaa:
Hey @cmajsmith, thanks for the pr! I'm trying to match the description with the PR code, could you please elaborate how this is working? Thanks.

nnixaa:
@cmajsmith ah, now I see it, thanks for the explanation! Is this a cross-browser way of fixing such issues?

cmajsmith:
@nnixaa, you're warmly welcome. I didn't test in all the browsers, but **Can I use** says: [Yes!](https://caniuse.com/#feat=css-gencontent)

nnixaa:
Hey @cmajsmith, got it. Though I personally feel like this is a bit fragile solution, and can be easily broken. Moreover, we have to introduce a new element in the dom, which is not particularly hidden.
What if we replace the `bold` font with a `text-shadow` instead? Like the second solution here https://www.sitepoint.com/quick-tip-fixing-font-weight-problem-hover-states/. We can even use the `currentColor` css value to not repeat the `color` in the shadow declaration. What do you think?

cmajsmith:
Hi, @nnixaa! 

I've considered using the solution you've provided before settling with the one I've applied. The problem with `text-shadow` is that it inevitably adds blur to the text and it looks rather ugly. 

To be honest, I think both (that `text-shadow` one and one in this PR) solutions are _hacky_. And thinking about the nature of the issue and ways to solve it I more and more get convinced that the best solution is... to get rid of bold styling for `:hover` state. 

Assuming the inactive and unhovered tab is pale and becomes darker on hover - the control response will be enough to be noticed. The shift of neighboring tabs on click (caused with active tab turning into bold) will still occur, but it won't be as noticeable as on `:hover`. The question here is: does it align with your team's design guidelines? If it is - that would be cleanest fix ever. What do you think?

nnixaa:
@cmajsmith I completely agree with you, let me check with the design team.

nnixaa:
Hey @cmajsmith, we discussed this and agreed that we can remove `bold` font-weight from the `:hover` state. However, we need to do this in both tabs components (tabset and route-tabset). Would you like to complete this PR considering this?

cmajsmith:
@nnixaa sure! I'll do it tomorrow once will have some time on my hands. Will update this PR once done.

cmajsmith:
@nnixaa, seems like that is it.

cmajsmith:
@nnixaa, it is still used for making active tab (both in tabset and route-tabset) bold. Bold was removed for hover only.

nnixaa:
@cmajsmith exactly! Merging then.

cmajsmith:
Hurray! Thanks @nnixaa for your time and help :)

Issue #653, opened 2018-08-20
NbAuthService removes properties from object


cloakedch:
### Issue type

**I'm submitting a ...**  (check one with "x")

* [x] bug report
* [ ] feature request

### Issue description

**Current behavior:**
Properties added to the object that is sent to the token endpoint are modified to (apparently) include only `email` and `password`.

**Expected behavior:**
Be able to include properties if desired by the user.

**Steps to reproduce:**
Auth0 working command:
`curl --request POST --url 'https://myendpoint.auth0.com/oauth/token' --header 'content-type: application/json' --data '{"grant_type":"password","username": "<clientID>","password": "<password>", "scope": "openid", "client_id": "<clientID>"}'`

Returns a fully working token. Note that `clientSecret` is not included and there is a field `username` instead of `email`.

Doing the same in Nebular using ngx-admin:

In `core.module.ts`:
```
NbOAuth2AuthStrategy.setup({
    name: 'myAuthStrategy',
    baseEndpoint: 'https://myendpoint.auth0.com',
    clientId: '<clientID>',
    clientSecret: 'none', // apparently necessary when using NbOAuth2ClientAuthMethod.REQUEST_BODY
    token: {
      endpoint: '/oauth/token',
      grantType: NbOAuth2GrantType.PASSWORD,
      class: NbAuthOAuth2Token
    },
    clientAuthMethod: NbOAuth2ClientAuthMethod.REQUEST_BODY
  })
```

Custom `login.component.ts` (Copied from framework source and showing only the method `login()`):
```
  login(): void {
    this.errors = this.messages = [];
    this.submitted = true;

    this.service.authenticate("myAuthStrategy", { username: 'some@test.com', password: 'test' }).subscribe((result: NbAuthResult) => {
      this.submitted = false;

      if (result.isSuccess()) {
        this.messages = result.getMessages();
      } else {
        this.errors = result.getErrors();
      }

      const redirect = result.getRedirect();
      if (redirect) {
        setTimeout(() => {
          return this.router.navigateByUrl(redirect);
        }, this.redirectDelay);
      }
      this.cd.detectChanges();
    });
  }
```

Looking in the developer console in chrome, the following reuqest body is sent:
![image](https://user-images.githubusercontent.com/32439069/44363315-2a98c380-a4c4-11e8-8873-577073f5bd13.png)

Note that `username` is missing from the body. If the property `username` was changed to `email`, then the request holds the specified properties.

alain-charles:
[...]

PR #659, opened 2018-08-22
NbOAuth2AuthStrategy token request body now provides username instead of email parameter

alain-charles:
According to [RFC6749 section 4.3.2](https://tools.ietf.org/html/rfc6749#section-4.3.2), the oAuth2  request body with `grant-type='password'` must provide `username` to the auth server's token endpoint  instead of `email`.

According to [RFC6749 section 4.3.2](https://tools.ietf.org/html/rfc6749#section-4.3.2), the oAuth2  request body with `grant-type='password'` may provide `scope` to the auth server's token endpoint.

=> Corrects issue #653


alain-charles:
And I missed it before 🤔

PR #687, opened 2018-08-28
fix(auth): fix isAuthenticatedOrRefresh for empty tokens

minextu:

`isAuthenticatedOrRefresh()` fails if no token exists: 
```
ERROR TypeError: There is no Auth Strategy registered under '' name
```

This is because token.getOwnerStrategyName() will also be empty if there never was a token or it has been deleted by logout. This will fix that.

### DIFF
--- a/src/framework/auth/services/auth.service.ts
+++ b/src/framework/auth/services/auth.service.ts
@@ -51,7 +51,7 @@ export class NbAuthService {
     return this.getToken()
       .pipe(
         switchMap(token => {
-        if (!token.isValid()) {
+        if (!token.isValid() && token.getOwnerStrategyName()) {
           return this.refreshToken(token.getOwnerStrategyName(), token)
             .pipe(
               switchMap(res => {
--- a/src/framework/auth/services/auth.spec.ts
+++ b/src/framework/auth/services/auth.spec.ts
@@ -31,7 +31,8 @@ describe('auth-service', () => {
   const resp200 = new HttpResponse<Object>({body: {}, status: 200});
 
   const testToken = nbAuthCreateToken(NbAuthSimpleToken, testTokenValue, ownerStrategyName);
-  const emptyToken = nbAuthCreateToken(NbAuthSimpleToken, null, ownerStrategyName);
+  const invalidToken = nbAuthCreateToken(NbAuthSimpleToken, null, ownerStrategyName);
+  const emptyToken = nbAuthCreateToken(NbAuthSimpleToken, null, null);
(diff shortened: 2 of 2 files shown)


alain-charles:
Hi @minextu,
I agree we can protect the call to `token.getStrategyName()` so that `if === ''`, the app will not fail.
But i think we have to understand why the token coming back from the localstorage has empty `ownerStrategyName`.

Please can you precise the steps to reproduce ? I tried opening two tabs in chrome, then logout in one of the tabs, then tried to call protected API on the other tab. In that case nothing happens (api is not called) but i did not see any error message in the console.

Thanks for your response

minextu:
I created a stackblitz: https://stackblitz.com/edit/angular-lhlrnu (folder auth was copied over from master)

It's pretty easy to reproduce: Just run `isAuthenticatedOrRefresh()` while not being logged in. `NbAuthJWTInterceptor` will do this automatically when requesting a protected API.

Thanks!

alain-charles:
OK @minextu .
So when we logout, i verified that we remove the token from storage, and we do.

So we have to work on `authService.getToken()`and the method it calls, because when there is no token in the localstorage, it stills creates that : 

capture d ecran 2018-08-29 a 10 42 52

>
Thanks for having found this issue.

alain-charles:
@nnixaa 

It seems that by construction, when the team wrote the token-parceler, it was important that even if there is no token in the localstorage, it stills does not return `null` value but a token with empty attributes.

This causes the above issue.

- Should we keep testing `token.getownerStrategyName()` before calling strategy's refreshToken (this PR)?
- Should we test instead `token.getValue()` ? (this makes more sense but it is extactly the same result)
- Should we refactor the token parceler so that it returns null value when no token in the storage ? In that case, we will have to pay attention to onTokenChange and token behaviourSubject if null values are filtered out.

What do you think ? I prepared a version with the quick fix consisting of testing the `token.getValue()`.

nnixaa:
@alain-charles, @minextu, sorry I'm out of town till Monday, will get back then.

nnixaa:
Hey @alain-charles, thanks for digging in.
The idea behind was that if you have something invalid in localstorage, the app won't crash and still will return you a token which is not valid. That's why the parceler has that `fallbackToken` thing.
In this particular case we are looking for an invalid token, but containing a previous value, so I agree with you that your second suggestion would be better rather than checking the owner name.
I will push a fix soon.

nnixaa:
Could you check the #708?

PR #708, opened 2018-09-11
fix(auth): fix `isAuthenticatedOrRefresh` to not refresh token with n…

nnixaa:
fix for https://github.com/akveo/nebular/pull/687

### DIFF
--- a/src/framework/auth/services/auth.service.ts
+++ b/src/framework/auth/services/auth.service.ts
@@ -51,7 +51,7 @@ export class NbAuthService {
     return this.getToken()
       .pipe(
         switchMap(token => {
-        if (!token.isValid()) {
+        if (!token.isValid() && token.getValue()) {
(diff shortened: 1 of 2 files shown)


src/framework/auth/services/auth.service.ts
```diff
@@ -51,7 +51,7 @@ export class NbAuthService {
     return this.getToken()
       .pipe(
         switchMap(token => {
-        if (!token.isValid()) {
+        if (!token.isValid() && token.getValue()) {
```
alain-charles:
```
if (token.getValue() && !token.isValid())
```
makes more sense according to me (and maybe better performance ?)

alain-charles:
@nnixaa i agree with the code you wrote. 
I only would test `token.getValue()` *and if truthy* `!token.isValid()` because i think is makes more sense.
```
if (token.getValue() && !token.isValid())
```

PR #712, opened 2018-09-13
Calendar localization update

pinckrow:

I added localization of dates and years in calendar components via adding new method getNumeric into NbLocaleService. Now it returns received values and all behaves as is. But this service can be overwritten with new implementation, returning regular numerics or special arabian characters based on selected locale.


Tibing:
Hi @pinckrow, we've removed `NbLocaleService` in [this PR](https://github.com/akveo/nebular/pull/732) and now all operations with dates performed through `NbDateService` which may be overwritten.
So, I think, your PR is no longer relevant.

PR #718, opened 2018-09-14
Hyperlink in action component

jalescardoso:
Currently the "nb-action" component does not support setting hyper links.
I'm adding support for angular router links and html links to the "NbActionsModule" module.

I did not set up the desktop, just imported the component and made the necessary customizations. So the compilation did not pass tests for simple deviations.
The team will have to configure this part to push the changes to production.
I did not find it necessary to create the complete desktop because they were simple changes.
Thank you


yggg:
Implemented in #1046

PR #724, opened 2018-09-18
fix(menu): item click now just Subject, no ReplaySubject

### LINKED ISSUE #695: `OnItemClick` stored in `ReplaySubject` cause logout right after login
### Issue type

* [x] bug report
* [ ] feature request

### Issue description

**Current behavior:**
@nnixaa I am subscribing to `onItemClick` then logout from the logout menu item is selected.
```
this.menuService.onItemClick()
    .pipe(takeUntil(this.ngUnsubscribe))
    .subscribe(item => {
      if (item.item.title === this.logoutMenuItem.title) {
        this.logout();
      }
    });
```
it successfully logs but on login again the `onItemClick` is triggered therefore logs out 
user immediately. That's because from the code the clicked items are stored in `ReplaySubject`
``` 
const itemClick$ = new ReplaySubject<NbMenuBag>(1);
```

**Expected behavior:**
i want on logout the `ReplaySubject` cache to be cleared out.

**Steps to reproduce:**
May be add an option to pass a null value as the last clicked item, so that the user has
an option to pass null menu item. so when the subscription picks up the replay it won't do anything. 

**Related code:**

### Other information:
**Angular, Nebular**
```
    "@nebular/theme": "^2.0.0-rc.9"
```

Tibing:

BREAKING CHANGE:

The `NbMenuService` not reply the last click event.

To Update:
if you use the knowledge that the last click event is replied you can wrap `onItemClick` stream in the custom stream based on `ReplaySubject`.

Closes #695


nnixaa:
@Tibing do we need to test it in any way? Are you sure it is not breaking anything?

PR #726, opened 2018-09-19
Changed OnInit to AfterViewInit for accordion-item-body component.

Ballpin:
This will give the component a chance to get the right height when
elements are being retrieved from REST API's or when other components
are involved.

### DIFF
--- a/src/framework/theme/components/accordion/accordion-item-body.component.ts
+++ b/src/framework/theme/components/accordion/accordion-item-body.component.ts
@@ -10,8 +10,7 @@ import {
   Host,
   ElementRef,
   ChangeDetectorRef,
-  OnInit,
-  OnDestroy,
+  OnDestroy, AfterViewInit,
 } from '@angular/core';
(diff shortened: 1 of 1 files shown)


nnixaa:
Hi @Ballpin, thanks for the PR! I presume this won't work in case when the content is dynamically changed in runtime? Should we change it to `onChanges`?

Ballpin:
Thank you @nnixaa for being of the devs creating nebular it's making wonders for me. :-)

I believe you are right on that part. For me it worked very well when a resolver was involved and prefetched the content. I could even use `<nb user></nb-user>` inside the accordion. A few hours later after submitting this PR, I noticed when you click at the "item-header" and expect the new content, it sticks to 35px height.

Ballpin:
Hi @nnixaa i tried to do it for several hours yesterday without success. Unfortunately i ended up with removing the animation instead for my use case.

If you succeed it would be awesome. :)

nnixaa:
Fixed in https://github.com/akveo/nebular/pull/882

PR #904, opened 2018-10-15
fix(theme): align context menu by center

### LINKED ISSUE #452: Not centered context menu
### Issue type

* [x] bug report
* [ ] feature request

### Issue description

**Current behavior:**

When creating a `<nb-user>` component only showing the picture (`onlyPicture=true`), the corresponding context menu is not centered under the picture. It looks like, even without showing the title or name of the user, a `div` with the class `info-container` is created which has a `margin-left: .5rem` attached. Without it, it works as expected.

**Expected behavior:**

Context menu is centered under the picture, even without showing the name and title of the user.

**Steps to reproduce:**

In your HTML code insert:
`<nb-actions>
          <nb-action>
            <nb-user onlyPicture="true" [nbContextMenu]="[{ title: 'Profile' }, { title: 'Log out'  }]" picture="http://gravatar.com/avatar"></nb-user>
          </nb-action>
        </nb-actions>`

**Related code:**

image

### Other information:

Node: v8.11.2
OS: Windows 10 Pro
Browser: Chrome

**Angular, Nebular**

Angular: 6.0.2
nebular/theme: 2.0.0-rc.8

lexzhukov:

Fix #452

### DIFF
--- a/src/framework/theme/components/user/user.component.html
+++ b/src/framework/theme/components/user/user.component.html
@@ -9,7 +9,7 @@
     <nb-badge *ngIf="badgeText" [text]="badgeText" [status]="badgeStatus" [position]="badgePosition"></nb-badge>
   </div>
 
-  <div class="info-container">
+  <div class="info-container" *ngIf="!onlyPictureMode">
     <div *ngIf="showNameValue && name" class="user-name">{{ name }}</div>
     <div *ngIf="showTitleValue && title" class="user-title">{{ title }}</div>
   </div>
--- a/src/framework/theme/components/user/user.component.ts
+++ b/src/framework/theme/components/user/user.component.ts
@@ -199,6 +199,10 @@ export class NbUserComponent {
   showInitialsValue: boolean = true;
   isMenuShown: boolean = false;
 
+  get onlyPictureMode(): boolean {
+    return !(this.showNameValue || this.showTitleValue);
+  }
+
   constructor(private domSanitizer: DomSanitizer) { }
 
   getInitials(): string {


src/framework/theme/components/user/user.component.html
```diff
@@ -9,7 +9,7 @@
     <nb-badge *ngIf="badgeText" [text]="badgeText" [status]="badgeStatus" [position]="badgePosition"></nb-badge>
   </div>
 
-  <div class="info-container">
+  <div class="info-container" *ngIf="!onlyPictureValue">
```
Tibing: Do we really need to introduce new variable here? Maybe, we'll just add this if?
```
  <div class="info-container" *ngIf="showNameValue || showTitleValue">
```

Tibing: @lexzhukov, What do you think?

lexzhukov: I've thought for your example. I don't like "mode" behavior without a special flag, but we can use property like `onlyPictureMode = showNameValue || showTitleValue`.

Tibing: I see it 😄 
But could you please explain why introducing new variable is better than using existing?

lexzhukov: It's explicit behavior in one word. I don't want to spend time on understanding multiple variables in one row:)

Tibing: Ok, It's clear enough, so, let's create a getter instead

src/framework/theme/components/user/user.component.ts
```diff
@@ -199,6 +199,10 @@ export class NbUserComponent {
   showInitialsValue: boolean = true;
   isMenuShown: boolean = false;
 
+  get onlyPictureMode(): boolean {
+    return this.showNameValue || this.showTitleValue;
+  }
```
Tibing: This getter lies me 😸, `onlyPictureMode` has to be true while we shouldn't `showNameValue` and `showTitleValue`. But now it does the inverse thing - it's true if we have to show either name or title or even both.
I think you have to add negation in the getter. Otherwise, it'll keep on lying.

lexzhukov: You're right. My bad.

src/framework/theme/components/user/user.component.ts
```diff
@@ -199,6 +199,10 @@ export class NbUserComponent {
   showInitialsValue: boolean = true;
   isMenuShown: boolean = false;
 
+  get onlyPictureMode(): boolean {
```
Tibing: Considering that this getter is used with negation in the template I would personally prefer to rename it in the reverse version like: `shouldShowUserInfo` or something similar. This way we may remove negations in the template and in the function's code.
What do you think?

Tibing: Also, what about testing?

Tibing: Considering that this getter is used with negation in the template I would personally prefer to rename it in the reverse version like: `shouldShowUserInfo` or something similar. This way we may remove negations in the template and in the function's code.
What do you think?

Tibing: Also, what about testing?

nnixaa:
Closing as updated

PR #1013, opened 2018-11-22
update analytics service

denStrigo:

### DIFF
--- a/docs/app/@theme/services/analytics.service.ts
+++ b/docs/app/@theme/services/analytics.service.ts
@@ -31,8 +31,14 @@ export class NgdAnalytics {
   }
 
   trackEvent(eventName: string, eventVal: string = '') {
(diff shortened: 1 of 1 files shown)


docs/app/@theme/services/analytics.service.ts
```diff
@@ -31,9 +31,15 @@ export class NgdAnalytics {
   }
 
   trackEvent(eventName: string, eventVal: string = '') {
-    if (this.enabled) {
+    if (this.enabled && this.window.ga) {
       ga('send', 'event', eventName, eventVal);
     }
+
+    if (!this.window.ga) {
```
nnixaa: it is mostly fine, but let's separate the `check` logic from the `track` method.
moreover, we also have the `trackPage` method which should also be wrapped into the `check` logic.

PR #1038, opened 2018-12-03
Added search suggestion capabilities for nbsearch

WLun001:
`input` event will trigger on every keystroke, therefore other developer can handle it and provide search suggestion as the user is typing.

Demo on `http://localhost:4200/#/search/search-event.component`

Added search suggestion capabilities, that will emit event on every keystroke. 
This can be handle by developers like this: https://angular.io/tutorial/toh-pt6#search-by-name


WLun001:
Hi @yggg 

Do you have any idea where should the search suggestion to be displayed if other developer used it? 
Currently the animated searches are taking whole screen, if the suggestion appears on their page (for example, layout rotate search will  push their page downwards), would be kinda weird.

yggg:
Hi @WLun001. Unfortunately, there is no way to customize search screen. 
As a workaround, you can create custom search components by extending `NbSearchFieldComponent` and `NbSearchComponent` and copying their templates. Then modify `NbSearchFieldComponent` template to have not only input but also search suggestions. Something like this:
```
@Component({
  selector: 'custom-search-field',
  // copied and modified template of NbSearchFieldComponent
  template: `
    ...
      <input #searchInput />
      <ul #options>...</ul>
    ...
  `,
})
class CustomSearchFieldComponent extends NbSearchFieldComponent {}

@Component({
  selector: 'custom-search-field',
  // template of NbSearchComponent
  template: `...`,
})
class CustomSearchComponentComponent extends NbSearchComponent {}
```

src/framework/theme/components/search/search.component.ts
```diff
@@ -149,6 +151,12 @@ export class NbSearchFieldComponent implements OnChanges, AfterViewInit {
     }
   }
 
+  submitSearchOnInput(term: string) {
+    if (term) {
```
yggg: It's better to behave like native input event, to not confuse people already familiar with it. So if native event emits when the value becomes empty, we should do this as well.

src/framework/theme/components/search/search.component.ts
```diff
@@ -320,6 +329,10 @@ export class NbSearchComponent implements OnInit, OnDestroy {
     this.hideSearch();
   }
 
+  searchOnInput(term: string) {
```
yggg: Let's change `searchOnInput` to `emitInput` to be more descriptive about what this method actually does.

src/framework/theme/components/search/search.component.ts
```diff
@@ -320,6 +329,10 @@ export class NbSearchComponent implements OnInit, OnDestroy {
...
+    this.searchService.submitSearch(term, this.tag);
```
yggg: `input` event actually more like change than submit, so I don't think we should trigger submit here. It would be better to also add `searchInput` and `onSearchInput` methods to `NbSearchService`.

src/playground/without-layout/search/search-with-input-event.component.ts
```diff
@@ -0,0 +1,42 @@
...
+  constructor(private searchService: NbSearchService) {
+  }
+
+  ngOnInit() {
+    this.searchService.onSearchSubmit().pipe(
+      debounceTime(300),
+      distinctUntilChanged(),
+    ).subscribe((data: { term: string, tag: string }) => {
+      this.value = data.term;
+      console.info(`term: ${data.term}, from search: ${data.tag}`);
```
yggg: Please, remove `console.log`.

src/framework/theme/components/search/search.component.ts
```diff
@@ -83,6 +84,7 @@ export class NbSearchFieldComponent implements OnChanges, AfterViewInit {
 
   @Output() close = new EventEmitter();
   @Output() search = new EventEmitter();
+  @Output() searchOnInput = new EventEmitter();
```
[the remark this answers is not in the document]
WLun001: For this case, can I renamed to `inputSearch` instead?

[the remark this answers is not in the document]
WLun001: @yggg if you are okay with this then I will commit my changes.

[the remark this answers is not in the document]
yggg: >The `input` event which get trigger is the Native Input Event, not the custom event emitted by `nb-search-field` as the name clashes

Oh, you're right. Sorry. Let it be `searchInput` then. 🙂

src/playground/without-layout/search/search-with-input-event.component.ts
```diff
@@ -0,0 +1,41 @@
...
+      <nb-layout-column>
+        <nb-card accent="info">
+          <nb-card-header>You searched for:</nb-card-header>
+          <nb-card-body>
+            <h3>{{ value || '-' }}</h3>
+          </nb-card-body>
+        </nb-card>
+      </nb-layout-column>
+    </nb-layout>
+
```
nnixaa: unnecessary empty line

yggg: Removed in f5d96dc8807911f3a5d617fdf4c768042bbe5aac

PR #1127, opened 2019-01-07
edit nbWindowFactory in theme.module.ts for SSR

alikarimii:
In SSR throw error window not define.when edit this function my problem resolved

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dev": true,
       "requires": {
         "@angular-devkit/core": "0.6.3",
-        "rxjs": "6.1.0"
+        "rxjs": "^6.0.0"
       },
       "dependencies": {
         "@angular-devkit/core": {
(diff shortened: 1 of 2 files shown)


nnixaa:
@yggg I guess this is unnecessary is should be resolved by providing a separate window object for SSR?

yggg:
@nnixaa `window` shouldn't be available on the server. Probably, we need to decorate all `window` and `document` injections with `Optional` and add null checks.

yggg:
Hi @alikarimii, thanks for the reporting!
I think we shouldn't modify window factory function since `window` doesn't exist on the server and changing `undefined` to `{}` doesn't change much. Any method call will throw anyways.

Angular shouldn't throw if factory function returns `undefined`. Probably, `window` injected and used somewhere in the code and since it's `undefined` error is thrown.
Check the error's call stack and you should see which component, service or directive tries to access window api. If it's coming from nebular component, please submit an [issue](https://github.com/akveo/nebular/issues/new).
If it's somewhere in our code, [here](https://angular.io/guide/universal#working-around-the-browser-apis) you can learn how to avoid it.

nnixaa:
Closing for now.

alikarimii:
base on this [repo](https://github.com/alikarimii/angular-universal-nebular-theme) i use nebular theme.
but when don't used `domino` on `server.ts` , throw error on server that window undefined and render is not completed.
when i rewrite nbWindowFactory(return window ? window : {}) every thing ok.

PR #1153, opened 2019-01-14
fix(popover): fix template context not being passed, update cdk to 7.2.1

### LINKED ISSUE #848: Refactor overlays after @angular/cdk@7.0.0 relase
We've added some workarounds in current overlays implementation because of incorrect implementation of cdk overlays. They don't provide the capability to provide component factory resolver into the portal, so we don't have the capability to render components in lazily loaded modules through services provided in the application root.

### LINKED ISSUE #1084: Popover with template ref doesn't render passed context
Check [this example](https://akveo.github.io/nebular/?p=/example/popover/popover-template-ref.component)

nnixaa:
Closes #1084 
Closes #848

### DIFF
--- a/src/framework/theme/components/cdk/overlay/mapping.ts
+++ b/src/framework/theme/components/cdk/overlay/mapping.ts
@@ -1,9 +1,7 @@
 import {
   ApplicationRef,
   ComponentFactoryResolver,
-  ComponentRef,
   Directive,
-  EmbeddedViewRef,
   Inject,
   Injectable,
   Injector,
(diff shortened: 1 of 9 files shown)


src/framework/theme/components/cdk/overlay/mapping.ts
```diff
@@ -41,9 +32,6 @@ import {
   ScrollStrategyOptions,
 } from '@angular/cdk/overlay';
 import { Platform } from '@angular/cdk/platform';
-import { Directionality } from '@angular/cdk/bidi';
```
Tibing: Please, also, remove this module from `rollup-config.ts`

PR #1223, opened 2019-02-10
Add compareWith parameter for Select Component

jgasparetti:

Add compareWith Input parameter in order to provide custom comparing Algorithm for the select Component


nnixaa:
Hey @jgasparetti, it seems that `looseIdentical ` is an internal API that we are not supposed to use. Is there any other way to achieve this?

jgasparetti:
Hi @nnixaa, I've remove the the `looseIdentical` in favor of a function that check for strict equalities

src/framework/theme/components/select/select.component.ts
```diff
@@ -198,6 +198,11 @@ export class NbSelectComponent<T> implements OnInit, AfterViewInit, AfterContent
    * */
   @Input() placeholder: string = '';
 
+  /**
+   * Custom comparing algorithm
+   */
+  @Input() compareWith: (o1: any, o2: any) => boolean = this.areIdentical;
```
yggg: Please, specify type to prevent accidental call with arguments other than value type. Like this:
`(o1: T, o2: T) => boolean`.

Also, what do you think about adding setter to compareWith to prevent setting values other than functions?

src/framework/theme/components/select/select.component.ts
```diff
@@ -422,7 +434,7 @@ export class NbSelectComponent<T> implements OnInit, AfterViewInit, AfterContent
   protected handleSingleSelect(option: NbOptionComponent<T>) {
     const selected = this.selectionModel.pop();
 
-    if (selected && selected !== option) {
+    if (selected && !this.compareWith(selected, option)) {
```
yggg: `selected` and `option` is instances of NbOptionComponent component. We can't compare them using possibly user defined comparator, since it expects to compare option values which type is `T` not `NbOptionComponent`.
Please, pass options values to comparator here.

PR #1254, opened 2019-02-21
refactor(tree-grid): allow specify getters for node properties

yggg:

Custom getters for node properties

### DIFF
--- a/src/framework/theme/components/tree-grid/data-source/tree-grid-data-source.ts
+++ b/src/framework/theme/components/tree-grid/data-source/tree-grid-data-source.ts
@@ -49,7 +49,7 @@ export class NbTreeGridDataSource<T> extends NbDataSource<NbTreeGridPresentation
     data: N[],
     dataGetter?: DataGetter<N, T>,
     childrenGetter?: ChildrenGetter<N, T>,
-    expandedGetter?: ExpandedGetter<N, T>,
+    expandedGetter?: ExpandedGetter<T>,
   ) {
     const presentationData: NbTreeGridPresentationNode<T>[] = data
       ? this.treeGridDataService.toPresentationNodes(data, dataGetter, childrenGetter, expandedGetter)
@@ -144,7 +144,7 @@ export class NbTreeGridDataSourceBuilder<T> {
     data: N[],
     dataGetter?: DataGetter<N, T>,
     childrenGetter?: ChildrenGetter<N, T>,
-    expandedGetter?: ExpandedGetter<N, T>,
+    expandedGetter?: ExpandedGetter<T>,
   ): NbTreeGridDataSource<T> {
     const dataSource = new NbTreeGridDataSource<T>(
       this.sortService,
--- a/src/framework/theme/components/tree-grid/data-source/tree-grid-data-source.ts
+++ b/src/framework/theme/components/tree-grid/data-source/tree-grid-data-source.ts
@@ -8,14 +8,19 @@ import { Injectable } from '@angular/core';
 import { BehaviorSubject, combineLatest, Observable } from 'rxjs';
 import { map } from 'rxjs/operators';
 
-import { NbDataSource } from '../../cdk/table';
 import { NbCollectionViewer } from '../../cdk/collections';
-import { NbTreeGridSortService } from './tree-grid-sort.service';
-import { NbTreeGridFilterService } from './tree-grid-filter.service';
-import { NbToggleOptions, NbTreeGridService } from './tree-grid.service';
-import { NbTreeGridDataService } from './tree-grid-data.service';
+import { NbDataSource } from '../../cdk/table';
 import { NbSortable, NbSortRequest } from '../tree-grid-sort.component';
-import { DEFAULT_ROW_LEVEL, NbTreeGridNode, NbTreeGridPresentationNode } from './tree-grid.model';
+import { NbTreeGridDataService } from './tree-grid-data.service';
+import { NbTreeGridFilterService } from './tree-grid-filter.service';
+import { NbTreeGridSortService } from './tree-grid-sort.service';
+import {
+  ChildrenGetter,
+  DataGetter,
+  DEFAULT_ROW_LEVEL, ExpandedGetter,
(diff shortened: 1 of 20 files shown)


src/framework/theme/components/tree-grid/data-source/tree-grid-data-source.ts
```diff
@@ -130,15 +140,20 @@ export class NbTreeGridDataSourceBuilder<T> {
               private treeGridDataService: NbTreeGridDataService<T>) {
   }
 
-  create(data: NbTreeGridNode<T>[]): NbTreeGridDataSource<T> {
+  create<N>(
+    data: N[],
+    dataGetter?: DataGetter<N, T>,
```
Tibing: Let's configure all these getters through an optional object with optional fields. It would be much more convenient in cases when you want to provide *expandedGetter* only and don't want to provide others.

Tibing: Something like this:
```
  interface Getters {
    dataGetter?: DataGetter<N, T>,
    childrenGetter?: ChildrenGetter<N, T>,
    expandedGetter?: ExpandedGetter<T>,
  }

  create<N>(
    data: N[],
    getters?: Getters
  ): NbTreeGridDataSource<T> {
  }
```

src/framework/theme/components/tree-grid/data-source/tree-grid-data.service.ts
```diff
@@ -3,22 +3,34 @@
...
-      const presentationNode = new NbTreeGridPresentationNode(node, level);
-
-      if (node.children) {
-        presentationNode.children = this.toPresentationNodes(node.children, level + 1);
-      }
-
-      return presentationNode;
+  toPresentationNodes<N>(
+    nodes: N[],
+    dataGetter: DataGetter<N, T> = node => node.data,
```
Tibing: Also, let's move these default getters out of the function declaration. It's overcomplicated now, don't you think so?

src/framework/theme/components/tree-grid/data-source/tree-grid-data.service.ts
```diff
@@ -3,22 +3,34 @@
...
+    childrenGetter: ChildrenGetter<N, T> = d => d.children || undefined,
+    expandedGetter: ExpandedGetter<T> = d => d.expanded,
+    level: number = DEFAULT_ROW_LEVEL,
+  ): NbTreeGridPresentationNode<T>[] {
+    return nodes.map(node => {
+      const children = childrenGetter(node);
+      const presentationChildren: NbTreeGridPresentationNode<T>[] = children
```
Tibing:
```suggestion
      const presentationChildren: NbTreeGridPresentationNode<T>[] = children && this.toPresentationNodes(children, dataGetter, childrenGetter, expandedGetter, level + 1) 
```
Don't you think it would be sligtly cleaner? 😄

src/framework/theme/components/tree-grid/data-source/tree-grid-data.service.ts
```diff
@@ -36,14 +48,11 @@ export class NbTreeGridDataService<T> {
 
   copy(nodes: NbTreeGridPresentationNode<T>[]): NbTreeGridPresentationNode<T>[] {
     return nodes.map((node: NbTreeGridPresentationNode<T>) => {
-      const presentationNode = new NbTreeGridPresentationNode(node.node, node.level);
-      presentationNode.expanded = node.expanded;
-
-      if (node.hasChildren()) {
-        presentationNode.children = this.copy(node.children);
-      }
+      const children: NbTreeGridPresentationNode<T>[] = node.hasChildren()
```
Tibing: The same thing here:
```suggestion
      const children: NbTreeGridPresentationNode<T>[] = node.hasChildren() && this.copy(node.children);
```

Why don't you like this style? It's not first I'm noticing you're using ternary operators instead of plain *&&*. Why do you decide that way?)

yggg: In case when `node.hasChildren()` is falsey value, children would become this value instead of `undefined`. Personally, I don't like this behavior, because if my property defined as `{} | undefined` I can use `value === undefined` to verify it's not defined. It won't work with other falsey values 🙂

src/framework/theme/components/tree-grid/data-source/tree-grid-data-source.ts
yggg: Yeah, that's much better. Thanks!

src/framework/theme/components/tree-grid/data-source/tree-grid-data.service.ts
Tibing: Ok, then, let's use plain *if/else* statement. I think it's the best solution 😄

yggg: Yep, did like this already. Thx

src/framework/theme/components/tree-grid/data-source/tree-grid.model.ts
```diff
@@ -4,52 +4,33 @@
...
-   * Child rows
-   */
-  children?: NbTreeGridNode<T>[];
-  /**
-   * Row expand state
-   */
-  expanded?: boolean;
-}
-
 export const DEFAULT_ROW_LEVEL: number = 0;
```
Tibing: The last thing. Let's add *Nb* prefix for *DEFAULT_ROW_LEVEL*, *DataGetter*, *ChildrenGetter*, *ExpandedGetter* and *Getters*.

yggg: Done in https://github.com/akveo/nebular/pull/1254/commits/e51c46afd5d28a81d8e9505021084bd5780ec004. Thanks!

src/framework/theme/components/tree-grid/data-source/tree-grid-data.service.ts
yggg: Done

PR #1256, opened 2019-02-21
refactor(theme): remove use of scss-functions

### LINKED ISSUE #1228: Remove scss functions from Nebular
- as the first step on a way to css-variables.

nnixaa:
BREAKING CHANGES:
- calendar - use primary button in cosmic theme
- checkbox - `opacity` instead of `lightning`
- context-menu, popover, datepicker - `calc` instead of `round` and scss calculations
- input - `opacity` for placeholder instead of `lightning`
- tabs - remove gradient for tab bottom separator in cosmic theme
- toastr - use `background-color` instead of gradient in cosmic theme
- text colors are now used from success/primary/warning/etc colors
- add color palett instead of generating colors using scss-functions, example for `primary`
`color-primary-200`
`color-primary-300`
`color-primary-400`
`color-primary-600`
`color-primary-700`

and create-colors-palette() scss function that can be used during theme installation process to generate palette automatically

Closes #1228


src/framework/theme/styles/_theming.scss
```diff
@@ -54,6 +54,35 @@ $nb-themes-export: () !global;
...
+    color-success:  20deg,
+    color-info: -10deg,
+    color-warning: 10deg,
+    color-danger: -20deg,
+  );
+
+  @each $key, $degree in $color-keys {
+    $color: map-get($theme, $key);
+
+    $color-300: tint($color, 14%);
```
yggg: Shouldn't `tint`, `mix` etc. be removed as well? And I can't find usages of `create-colors-palette` function.

nnixaa: this function allows generating colors from existing ones. We need to add it to the migration guide https://github.com/akveo/nebular/issues/1309.

PR #1257, opened 2019-02-21
feat(theme): css variables support

### LINKED ISSUE #46: Support CSS Variables
I know this is on `theme/styles/_theming.scss` but setting it to true doesn't make `@angular/cli`'s webpack happy.
``` scss
$nb-enable-css-variables: false !global;
```
I would like to be able to use CSS variables in my projects!!
Thank you!

nnixaa:
Closes #46


src/framework/theme/styles/_theming.scss
```diff
@@ -101,14 +133,19 @@ $nb-themes-export: () !global;
...
     @each $var, $value in $variables {
-      --#{$var}: $value;
+      --#{$var}: #{$value};
     }
   }
 }
 
-// TODO: we hide :host inside of it which is not obvious
-@mixin nb-install-component() {
+@mixin nb-install-component-new() {
```
yggg: Maybe it's better to change `new` to `with-css-props-support`? Same for `old` and following `nb-install-global-new` and `nb-install-global-old` mixins.

src/framework/theme/styles/_theming.scss
```diff
@@ -13,7 +13,7 @@
 @import 'core/functions';
 
 $nb-enabled-themes: () !global;
-$nb-enable-css-variables: false !global;
+$nb-enable-css-variables: true !global !default;
```
yggg: Also maybe we should use `css-custom-properties` instead of `variables`. Thats how it called in [spec](https://www.w3.org/TR/css-variables-1/#custom-property). (Please, don't look at the url 🙂)

https://stackoverflow.com/questions/48873595/is-there-a-difference-between-css-custom-properties-and-css-variables

Issue #1260, opened 2019-02-22
nb-select doesn't deselect option when resetting ngModel


frankie567:
### Issue type

**I'm submitting a ...**  (check one with "x")

* [X] bug report
* [ ] feature request

### Issue description

**Current behavior:**

After having selected an option with `nb-select`, if the binded `ngModel` is resetted (`null` or `undefined` value) by the component, the option is still selected in the `nb-select`.

**Expected behavior:**

After having selected an option with `nb-select`, if the binded `ngModel` is resetted (`null` or `undefined` value) by the component, the option should be unselected in the `nb-select`.

**Steps to reproduce:**

* Define a data model in the component.
* Bind it to a `nb-select` in the template.
* Create a button that resets the data model to `null` on click.
* Select an option in `nb-select`.
* Click on the button.

**Related code:**

https://stackblitz.com/edit/github-naqwyy

### Other information:

* npm: 6.4.1
* node: 10.15.0
* macOS 10.14.2
* Chrome 72

**Angular, Nebular**

* @angular/core: 7.1.4
* @angular/forms: 7.1.4
* @nebular/theme: 3.3.0

aefox:
Fix provided in https://github.com/akveo/nebular/pull/1265 (will require discussion and if approved adding some tests)

aefox:
Fix provided in https://github.com/akveo/nebular/pull/1282

yggg:
Closing as fix (#1282) was merged

TeodorKolev:
This is not fixed at version 3.1.0 that we are using...

PR #1265, opened 2019-02-27
fix(select): select option not cleaned up when binding was updated programmatically

aefox:

This fixes https://github.com/akveo/nebular/issues/1260 where select was not being cleaned up when its binding was set to null programmatically.


aefox:
Please note that this PR breaks the guideline to ` Use changeDetection: ChangeDetectionStrategy.OnPush where applicable`. See it [here](https://github.com/akveo/nebular/pull/1265/files#diff-6d5a993596400c649c6d5287125c6acaL149).

Because of this I did not write tests since I first need to know this is acceptable.

src/framework/theme/components/select/select.component.ts
```diff
@@ -371,6 +369,7 @@ export class NbSelectComponent<T> implements OnInit, AfterViewInit, AfterContent
 
   writeValue(value: T | T[]): void {
     if (!value) {
+      this.cleanSelection();
```
yggg: Can't we keep `OnPush` and manually call `this.cd.detectChanges` here after selection cleanup?

package.json
```diff
@@ -174,4 +174,4 @@
     "uglifyjs-webpack-plugin": "1.1.5"
   },
   "schematics": "./schematics/dist/collection.json"
-}
\ No newline at end of file
+}
```
yggg: Please, revert whitespace/newline changes here and also in `package-lock.json` and `select-showcase.component.html`.

aefox:
@yggg thanks for the review!

I'm closing this as all changes from https://github.com/akveo/nebular/pull/1282 are also fixing this issue. I've tested all possible scenarios for this ticket and you can see the gist with the local changes [here](https://gist.github.com/aefox/8f3fd6b64d0c8ae3772ba69102e9bb33)

I will update https://github.com/akveo/nebular/pull/1282 to also mention this issue is covered by those changes.

Issue #973, opened 2018-11-11
NbContextMenu in Internet Explorer (bad styling)


Roman-Simik:
### Issue type

* Bug report

### Issue description

**Current behavior:**
NbContextMenu in Internet Explorer - **bad** context styling (position)

![image](https://user-images.githubusercontent.com/41468539/48316993-42777380-e5eb-11e8-83bc-6dcf9515d48f.png)

**Expected behavior:**
NbContextMenu in Internet Explorer - **good** context styling (position)

![image](https://user-images.githubusercontent.com/41468539/48316999-5de27e80-e5eb-11e8-9571-2edb5d21f9f3.png)

**Steps to reproduce:**
Just open the docs in Internet Explorer- https://akveo.github.io/nebular/docs/components/context-menu/overview#nbcontextmenudirective

### Other information:
Windows 10
Internet Explorer 11

aefox:
Fix for the arrow provided in https://github.com/akveo/nebular/pull/1268. Still trying to fix the menu items alignment as well ...

PR #1268, opened 2019-03-01
fix(context-menu): updated context menu arrow left position calculation...

aefox:
to take into account IE limitations

According to https://github.com/akveo/nebular/issues/973 the white part of the arrow is misplaced on IE. That is because the `left` property value upon rendering becomes something like `calc(50% - calc(11px - 2px))`. IE does not support this style and the calculation is now changed to become `calc(50% - (11px - 2px))`.

Two extra things to note:
- Initially I tried to make it be `calc(50% - 11px - 2px)` but for some very weird reason the result of that is incorrect (meaning the number of pixels is incorrect and the white part of the arrow is still misplaced
- I tried for several hours to also re-align the menu items contents (as you can see in https://github.com/akveo/nebular/issues/973 they are misaligned as well) but to no avail. Found the problem though: the span which contains the actual `menu-item` text is 0


nnixaa:
Hey @aefox, thanks for digging in! I had a quick look and it seems that the `span` is missing unfolded flex declaration `flex: 1 0 auto`. Could you try and update the PR?

aefox:
@nnixaa brilliantly spot on! works like a charm on all browsers (IE 11, Edge, FF and Chrome)

nnixaa:
Great job @aefox! Looking forward to looking into your other PRs.

Issue #1063, opened 2018-12-11
Cannot select option when bind data dynamic into nb-select


tuannh2:
I'm using nb-select for my project, and when i bind data auto in nb-select, i cannot select option as below images:

![image](https://user-images.githubusercontent.com/33337972/49795169-e596e680-fd6b-11e8-8b0f-448a8fc656c6.png)

![image](https://user-images.githubusercontent.com/33337972/49795213-04957880-fd6c-11e8-80c3-8ec37bee9d80.png)

my code in component:

"
import { Component } from '@angular/core';
import { NbMenuService } from '@nebular/theme';
import { TaxAuthorityService } from '../../@core/services/taxAuthority.service';

@Component({
    selector: 'ngx-organization',
    styleUrls: ['organization.component.scss'],
    templateUrl: 'organization.component.html',
})

export class OrganizationComponent {
    taxAuthorities;

    constructor(private menuService: NbMenuService, private taxAuthorityService: TaxAuthorityService) {
        //this.taxAuthorities = taxAuthorityService.getTaxAuthorities();
    }

    ngOnInit() {
        this.taxAuthorityService.getTaxAuthorities().subscribe(p=>this.taxAuthorities = p);
    }
}
"

Thanks!

Tuan,

gaspardip:
Its already fixed by #1009 but Nebular 3.0.1 is not complete yet

Duplicates: #948 #931

tanzeelrana:
is this fix released ?

tuannh2:
Thanks @tanzeelrana. 

I already used the normal select control instead of nb-select.

Tuan,

jaapcrezee:
This does not seem fixed in Nebular v3.3.0. When I set my .ts property to = null and use  data subscription and
```
  <nb-select *ngIf="items"
      placeholder="{{selector_input.placeholder}}"
  >
    <ng-container *ngFor="let item of items">
      <nb-option>{{ item }}</nb-option>
    </ng-container>
</nb-select>
```
there is no data in the dropdown select; <select/> works as expected.

nnixaa:
@Tibing can you check?

aefox:
[...]

PR #1280, opened 2019-03-06
fix(select): select not rendering option wrapped in ng container

aefox:

This fixes issue mentioned in https://github.com/akveo/nebular/issues/1063 related to `nb-select` not rendering `nb-options` that are wrapped in `ng-container`

### DIFF
--- a/src/framework/theme/components/select/select.component.html
+++ b/src/framework/theme/components/select/select.component.html
@@ -28,6 +28,6 @@
 
 <nb-card *nbPortal class="select" [ngClass]="[status, overlayPosition]" [style.width.px]="hostWidth">
   <nb-card-body>
-    <ng-content select="nb-option, nb-option-group"></ng-content>
+    <ng-content select="nb-option, nb-option-group, ng-container"></ng-content>
   </nb-card-body>
 </nb-card>
--- a/src/framework/theme/components/select/select.spec.ts
+++ b/src/framework/theme/components/select/select.spec.ts
@@ -225,7 +225,7 @@ describe('Component: NbSelectComponent', () => {
     })
   });
 
-  fit('should render option wrapped in ng-container', () => {
+  it('should render option wrapped in ng-container', () => {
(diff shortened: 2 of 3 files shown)


src/framework/theme/components/select/select.component.html
```diff
@@ -28,6 +28,6 @@
 
 <nb-card *nbPortal class="select" [ngClass]="[status, overlayPosition]" [style.width.px]="hostWidth">
   <nb-card-body>
-    <ng-content select="nb-option, nb-option-group"></ng-content>
+    <ng-content select="nb-option, nb-option-group, ng-container"></ng-content>
```
yggg: Not sure if it's ok to select `ng-container` here. This way it would be possible to pass arbitrary content into select. What do you think @aefox? Probably there is something we could do to restrict `ng-container` content (e.g. more specific selector)?

aefox: @yggg I totally see your point on `possible to pass arbitrary content into select`. That indeed is a problem I did not figure out, nice catch 👍 

At the same time, please note the following `hack/trick`:
```
<nb-select placeholder="Select Showcase" [(selected)]="selectedItem">
  <ng-container *ngFor="let item of items" ngProjectAs="nb-option">
    <div>{{ item.label }}</div>
  </ng-container>
</nb-select>
```

This is currently possible to do and it works. Works here means it actually renders the `div`s correctly (they ofc do not work as an `option`, but that is expected).

I'm not sure if we should force the users of `nebular` to restrict them from using `ng-container` in this situation as probably some devs will prefer to use `ng-container` for loops and keep their `nb-options` cleaner.

Indeed doing this we're opening up the possibility for miss usage, but the same can happen for the pure html tags (`select` and `option`).

I'm basically opened to doing this any way you think it is best.

nu5rim3:
any update?

JihadMotii-REISys:
Any updates?

Issue #1235, opened 2019-02-14
nb-select with empty value


SashaSkywalker:
### Issue type

**I'm submitting a ...**  (check one with "x")

* [x] bug report
* [ ] feature request

**Current behavior:**
nb-select render empty string in label for option with empty value

**Expected behavior:**
nb-select should render option label

**Steps to reproduce:**
```
<nb-select>
  <nb-option value=''>first</nb-option>
  <nb-option value='2'>second</nb-option>
</nb-select>
```

if user selects first option then nothing will be rendered in label (bad)
if user selects second option then *second* will be rendered in label (good)

rhajizadeh:
The same thing happens when with value equal to number 0 when we assign numbers to values.
for example:
```html
<nb-select>
  <nb-option [value]="0">
     first
   </nb-option>
   <nb-option [value]="1">
     test 2
  </nb-option>
</nb-select>
```
or: https://nb-select-zero-issue.stackblitz.io

aefox:
PR for fix opened in https://github.com/akveo/nebular/pull/1282

yggg:
Closing as fix (#1282) was merged

PR #1282, opened 2019-03-07
fix(select): allow select to work with empty values in nb-option

aefox:

This PR fixes https://github.com/akveo/nebular/issues/1235 and in order to allow `nb-options` have values of `""` or `"0"`.

These changes also end up fixes https://github.com/akveo/nebular/issues/1260.

BREAKING CHANGE:

Only 'null' and 'undefined' option values now considered as reset. `false` and falsy values such as `0`, `''`, `NaN` don't reset select value anymore.


src/framework/theme/components/select/select.component.ts
```diff
@@ -385,7 +381,7 @@ export class NbSelectComponent<T> implements OnInit, AfterViewInit, AfterContent
    * Selects option or clear all selected options if value is null.
    * */
   protected handleSelect(option: NbOptionComponent<T>) {
-    if (option.value) {
+    if (this.options.some((opt) => opt.content === option.content)) {
```
aefox: @nnixaa @Tibing I really don't like thi but couldn't find a better way. Basically, in order to know if we want to update the selection we search the existing options and try to match based on the `nb-option` content (which is its label). if you guys have a better idea I'm definitely opened.

PS: we could expose besides the `content` property a new `value` property which would tap into the native element's attributes and take the value from there, but we'd still have to do this kind of check in order to bypass javascript's falsy values

yggg: Probably we could use `option.value != null`. Loose comparison with `null` is true only for `null` and `undefined`, so it'll bypass any other falsy values.
@aefox Do you see any problems with this solution?

Could you also add small clarification to the select docs, please? Currently, it says 'Items without values will clean selection.' We also should mention `null` and `undefined`.

And this is a breaking change since behavior for people using falsy values (except `null` and `undefined`) will change.

aefox: @yggg thanks a lot for the review!
Indeed you are correct, loose comparison to null is definitely the way to go here, nicely spotted. Also, nice catch on the docs, updated those as well.

Not sure about the breaking changes. Should I do anything in particular for this?

yggg: >Not sure about the breaking changes. Should I do anything in particular for this?

Commit which introduce breaking change should has `BREAKING CHANGE:` with description inside commit message body. See details in [CONTRIBUTING.md](https://github.com/akveo/nebular/blob/master/CONTRIBUTING.md#footer). Please, update commit and force push.

It also means we would be able to merge this PR only in major version update (Nebular 4.x.x) since we follow the [semantic versioning spec](https://semver.org/).

src/framework/theme/components/select/select.component.ts
```diff
@@ -385,7 +381,7 @@ export class NbSelectComponent<T> implements OnInit, AfterViewInit, AfterContent
    * Selects option or clear all selected options if value is null.
    * */
   protected handleSelect(option: NbOptionComponent<T>) {
-    if (option.value) {
+    if (option.value != null) {
```
yggg: Could you please swap if and else bodies and use positive comparison (`==` instead `!=`)?
I mean
```
if (option.value == null) {
  this.reset();
} else {
  this.selectOption(option);
}
```
Sorry for bothering you with such changes, but this way it simplifies reading a lot.

src/playground/with-layout/select/select-clean.component.html
```diff
@@ -1,5 +1,6 @@
 <nb-select placeholder="Cleanable">
   <nb-option>Clean</nb-option>
+  <nb-option [value]="nullValue">Clean with null value</nb-option>
```
yggg: Why do you use property set to null instead of setting null in input directly?
`[value]="null"`
Also, what do you think about adding one more option without value, to illustrate another reset way to make reset option.

src/playground/with-layout/select/select-showcase.component.html
```diff
@@ -1,7 +1,8 @@
 <nb-select placeholder="Select Showcase" [(selected)]="selectedItem">
+  <nb-option value="">Option empty</nb-option>
+  <nb-option value="0">Option 0</nb-option>
   <nb-option value="1">Option 1</nb-option>
   <nb-option value="2">Option 2</nb-option>
   <nb-option value="3">Option 3</nb-option>
   <nb-option value="4">Option 4</nb-option>
 </nb-select>
-
```
yggg: Please revert new line change.
By the way, to maintain consistent code style and prevent such changes we have `.editorconfig` file in our repo. If you'd like, you could install [editorconfig plugin](https://editorconfig.org/#download) for your editor or IDE, so it will automatically follow rules defined in project config, such as new line at the end of file.

PR #1340, opened 2019-04-02
fix(select): prevent dispose call on uninitialized properties

yggg:

Fixes case when component was destroyed before calling init hook.

### DIFF
--- a/src/framework/theme/components/select/select.spec.ts
+++ b/src/framework/theme/components/select/select.spec.ts
@@ -281,6 +281,11 @@ describe('Component: NbSelectComponent', () => {
     expect(selectButton.textContent).toEqual(selectedOption.value.toString());
   }));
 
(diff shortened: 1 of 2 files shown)


src/framework/theme/components/select/select.spec.ts
```diff
@@ -281,6 +281,11 @@ describe('Component: NbSelectComponent', () => {
     expect(selectButton.textContent).toEqual(selectedOption.value.toString());
   }));
 
+  it(`should not call dispose on initialized resources`, () => {
```
nnixaa: on _un_ initialized?

yggg: Right, fixed in d8a2d4e409feddf1798a97ffec5895d6a1be8fe3.

PR #1380, opened 2019-04-12
fix(schematics): fix eva-icons schematic

nnixaa:


nnixaa:
BLOCKED: revert travis fix

PR #1416, opened 2019-04-24
fix(option group): propagate disabled state to child options

yggg:

Disable options if group disabled


src/framework/theme/components/select/option-group.component.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    fixture.detectChanges();
+
+    testComponent.optionGroupDisabled = false;
+    fixture.detectChanges();
+
+    const optionGroup = fixture.debugElement.query(By.directive(NbOptionGroupComponent));
+    expect(optionGroup.attributes.disabled).toEqual(null);
+  });
+
+  it('should disable options if set disabled', () => {
```
Tibing: I think that way test name will be more readable 😄 
```suggestion
  it('should disable group options if group disabled', () => {
```

src/framework/theme/components/select/option-group.component.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    const setDisabledSpy = spyOn(optionComponent, 'setDisabledByGroupState');
+
+    optionGroupComponent.disabled = true;
+    fixture.detectChanges();
+
+    expect(setDisabledSpy).toHaveBeenCalledTimes(1);
+    expect(setDisabledSpy).toHaveBeenCalledWith(true);
+  });
+
+  it('should enable options if set enabled', () => {
```
Tibing: I think that way test name will be more readable 😄 
```suggestion
  it('should enable group options if group enabled', () => {
```

src/framework/theme/components/select/option-group.component.ts
```diff
@@ -26,13 +40,49 @@ export class NbOptionGroupComponent {
   }
   set disabled(value: boolean) {
     this._disabled = convertToBoolProperty(value);
+
+    if (this.options) {
```
Tibing: Don't you think we have to check `this.options.length` here?
```suggestion
    if (this.options && this.options.length) {
```

src/framework/theme/components/select/option-group.component.ts
```diff
@@ -26,13 +40,49 @@ export class NbOptionGroupComponent {
...
+
+  @ContentChildren(NbOptionComponent, { descendants: true }) options: QueryList<NbOptionComponent<any>>;
+
+  ngAfterContentInit() {
+    if (this.options.length) {
+      this.asyncUpdateOptionsDisabledState();
+    }
+
+    this.options.changes
+      .pipe(takeWhile(() => this.alive))
```
Tibing: Maybe, let's start using `takeUntil(this.destroyed$)` 😄 in new components?

src/framework/theme/components/select/option-group.component.ts
```diff
@@ -26,13 +40,49 @@ export class NbOptionGroupComponent {
...
+
+  ngOnDestroy() {
+    this.alive = false;
+  }
+
+  /**
+   * Sets disabled state for each option to current group disabled state.
+   */
+  protected updateOptionsDisabledState(): void {
+    this.options.forEach((option: NbOptionComponent<any>) => option.setDisabledByGroupState(this.disabled));
```
Tibing: Are you sure we have to use `NbOptionComponent<any>` here? Why can't we make `NbOptionGroupComponent` also generic?

src/framework/theme/components/select/option-group.component.ts
```diff
@@ -26,13 +40,49 @@ export class NbOptionGroupComponent {
...
+  }
+
+  ngOnDestroy() {
+    this.alive = false;
+  }
+
+  /**
+   * Sets disabled state for each option to current group disabled state.
+   */
+  protected updateOptionsDisabledState(): void {
```
Tibing: I don't think we have to use `void` return statements :) Because we're not using it always. And I personally hate that thing 😈

src/framework/theme/components/select/option-group.component.ts
```diff
@@ -26,13 +40,49 @@ export class NbOptionGroupComponent {
...
+    this.options.forEach((option: NbOptionComponent<any>) => option.setDisabledByGroupState(this.disabled));
+  }
+
+  /**
+   * Updates options disabled state after promise resolution.
+   * This way change detection will be triggered after options state updated.
+   * Use this method when updating options during change detection run (e.g. QueryList.changes, lifecycle hooks).
+   */
+  protected asyncUpdateOptionsDisabledState(): void {
```
Tibing: I don't think we have to use `void` return statements :) Because we're not using it always. And I personally hate that thing 😈

src/framework/theme/components/select/option.component.ts
```diff
@@ -42,6 +42,9 @@ import { NbSelectComponent } from './select.component';
   `,
 })
 export class NbOptionComponent<T> implements OnDestroy {
+
+  protected disabledByGroup: boolean = false;
```
Tibing: We don't need to specify `: boolean` explicitly here. Because ts compiler can infer type when you assign value in declaration. Also as it's quite obvious that this value always will be `boolean` 😄 
As I remember we have the following `tslint` rule.

src/framework/theme/components/select/option-group.component.ts
```diff
@@ -17,7 +28,10 @@ import { convertToBoolProperty } from '../helpers';
     <ng-content select="nb-option, ng-container"></ng-content>
   `,
 })
-export class NbOptionGroupComponent {
+export class NbOptionGroupComponent implements AfterContentInit, OnDestroy {
+
+  protected alive: boolean = true;
```
Tibing: We don't need to specify `: boolean` explicitly here. Because ts compiler can infer type when you assign value in declaration. Also as it's quite obvious that this value always will be `boolean` 😄 
As I remember we have the following `tslint` rule.

src/framework/theme/components/select/option.component.ts
```diff
@@ -103,7 +106,7 @@ export class NbOptionComponent<T> implements OnDestroy {
 
   @HostBinding('attr.disabled')
   get disabledAttribute(): '' | null {
-    return this.disabled ? '' : null;
+    return (this.disabledByGroup || this.disabled) ? '' : null;
```
Tibing: Could you please split that line into multiple lines to make it more readable. Like so
```
const disabled = this.disabledByGroup || this.disabled;
return disabled ? '' : null;
```
Or even using if statement

src/framework/theme/components/select/option.component.ts
```diff
@@ -119,6 +122,17 @@ export class NbOptionComponent<T> implements OnDestroy {
     this.setSelection(false);
   }
 
+  /**
+   * Sets disabled by group state and marks component for check.
+   * @param disabled group disabled state
```
Tibing: Do we really need to specify that `@param`?

src/framework/theme/components/select/option.component.ts
```diff
@@ -119,6 +122,17 @@ export class NbOptionComponent<T> implements OnDestroy {
...
+   */
+  setDisabledByGroupState(disabled: boolean): void {
```
Tibing: Again, I'm confused with `void` 😄

src/framework/theme/components/select/option-group.component.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+            <nb-option *ngIf="showOption" [value]="1" [disabled]="optionDisabled">1</nb-option>
+          </nb-option-group>
+        </nb-select>
+
+      </nb-layout-column>
+    </nb-layout>
+  `,
+})
+export class NbOptionGroupTestComponent {
+  selectDisabled: boolean = false;
```
Tibing: Let's not specify the property type when we assign value during the declaration. I've described that slightly deeper in some comment below.

src/framework/theme/components/select/select.spec.ts
```diff
@@ -248,6 +249,30 @@ export class NbSelectWithFalsyOptionValuesComponent {
...
+            <nb-option [value]="1" [disabled]="optionDisabled">1</nb-option>
+          </nb-option-group>
+        </nb-select>
+
+      </nb-layout-column>
+    </nb-layout>
+  `,
+})
+export class NbOptionDisabledTestComponent {
+  optionGroupDisabled: boolean = false;
```
Tibing: Let's not specify the property type when we assign value during the declaration. I've described that slightly deeper in some comment above.

src/framework/theme/components/select/option-group.component.ts
yggg: I don't see why we should add additional check. In `updateOptionsDisabledState` we iterate through options and if there is no options, nothing will be done.

src/framework/theme/components/select/option-group.component.ts
yggg: I think we'll refactor components all together later and before that moment we should stick with a current agreed style. Consistency, you know. 🙂

src/framework/theme/components/select/option-group.component.ts
yggg: I like the idea, but then everyone who used NbOptionGroupComponent has to change their code. Not sure is it worth it since group has nothing to do with option values.

src/framework/theme/components/select/option-group.component.ts
yggg: It will prevent you from returning values if you try to. 🙂

src/framework/theme/components/select/option-group.component.ts
yggg: It will prevent you from returning values if you try to. 🙂

PR #1585, opened 2019-06-06
refactor(checkbox): deprecate `value`, add `checked`, don't emit of i…

### LINKED ISSUE #1478: nb-checkbox (change) event.returnValue is always true in FireFox.
### Issue type

**I'm submitting a ...

* [x] bug report
* [ ] feature request

### Issue description

**Current behavior:**
The (change) output event of nb-checkbox component does not provide the correct value for the checkbox in firefox, as it is always true. 

**Expected behavior:**
The (change) output event of nb-checkbox component should provide the correct value of the checkbox

**Steps to reproduce:**
Use the nb-checkbox component with a change output event. In the consuming method, the event.returnValue should be the value of the checkbox but is always true in FireFox. 

**Related code:**
https://stackblitz.com/edit/github-2uedts-mqc2hn

```
<nb-checkbox (change)="checkValue($event)">Show all issues</nb-checkbox>

checkValue(event: any) {
    console.log(event.returnValue)
  }
```

### Other information:

**Angular, Nebular**
```
@nebular/theme: 3.5.0
@angular: 7.*.*
```

nnixaa:
…nput change

Closes #1478

### DIFF
--- a/src/framework/theme/components/checkbox/checkbox.component.ts
+++ b/src/framework/theme/components/checkbox/checkbox.component.ts
@@ -179,7 +179,6 @@ export class NbCheckboxComponent implements ControlValueAccessor {
   }
   set checked(value: boolean) {
     this._checked = value;
-    this.onChange(value);
   }
   private _checked: boolean = false;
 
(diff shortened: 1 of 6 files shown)


src/framework/theme/components/checkbox/checkbox.component.ts
```diff
@@ -201,7 +214,11 @@ export class NbCheckboxComponent implements ControlValueAccessor {
   }
   private _indeterminate: boolean = false;
 
-  @Output() valueChange = new EventEmitter();
+  /**
+   * Output when value is changed by a user
+   * @type EventEmitter<boolean>
+   */
+  @Output() valueChange = new EventEmitter<boolean>();
```
yggg: I think we also need to deprecate `valueChange` and add `checkedChange`

PR #1658, opened 2019-06-29
fix(flip card): add prefixed backface-visibility property

yggg:

Fixes issue with Safari when back card appears on top of front card and
flip button making them unclickable.


src/framework/theme/components/card/flip-card/flip-card.component.scss
```diff
@@ -18,6 +18,9 @@
       opacity: 0;
       transition: opacity 0s 0.25s;
       backface-visibility: hidden;
+      // Don't remove prefixed property as it's wouldn't be added during the build.
```
Tibing: Don't you think it would be better to use some kind of mixin instead of copy-pasting the same code + comment?

yggg: I do 🙂
Fixed in 008f809

Tibing: Nice! LGTM

PR #1034, opened 2018-11-30
fix(menu): highlight when active anchors

Tibing:
### Please read and mark the following check list before creating a pull request:

 - [x] I read and followed the [CONTRIBUTING.md](https://github.com/akveo/nebular/blob/master/CONTRIBUTING.md) guide.
 - [x] I read and followed the [New Feature Checklist](https://github.com/akveo/nebular/blob/master/DEV_DOCS.md#new-feature-checklist) guide.
 
#### Short description of what this resolves:

Closes #875

PR #1660, opened 2019-06-29
docs(menu): Fix selection highlight in menu service example

miqh:
 
---

Hello,

This is a very minor documentation fix.

I just noticed the menu service example doesn't correctly highlight one of the last items in the menu, presumably due to how the menu item is configured to match (i.e. full path by default).

### DIFF
--- a/src/playground/with-layout/menu/menu-service-items.ts
+++ b/src/playground/with-layout/menu/menu-service-items.ts
@@ -39,8 +39,6 @@ export const MENU_ITEMS = [
             title: 'Commercial',
             link: '/example/menu/menu-service.component/3/3/2',
             icon: 'grid-outline',
-            queryParams: {param: 2},
-            fragment: 'fragment',
           },
         ],
       },


src/playground/with-layout/menu/menu-service-items.ts
```diff
@@ -39,8 +39,6 @@ export const MENU_ITEMS = [
             title: 'Commercial',
             link: '/example/menu/menu-service.component/3/3/2',
             icon: 'grid-outline',
```
yggg: I'm afraid there is no issue with configuration. Looks like the fix (#1034) aimed to include fragments into comparison doesn't work.

See here:
https://github.com/akveo/nebular/blob/f65a2cc7b0d589e6532e9705ea21f98312441bdf/src/framework/theme/components/menu/menu.service.ts#L387
There are calls for `isFragmentEqual` and `isFragmentContain` functions which expecting path with the fragment. But `this.location.path()` doesn't return fragment unless you provide `true` as a parameter to the `path` function.

So a proper fix for the issue would be adding `true` as a parameter for `this.location.path` calls on lines `387` and `388`.

Another thing to mention, there is plenty of tests for path and fragment comparison [here](https://github.com/akveo/nebular/blob/f65a2cc7b0d589e6532e9705ea21f98312441bdf/src/framework/theme/components/menu/menu.spec.ts#L303) and they all are passing. As turns out, `SpyLocation` used in unit tests as a mock for `Location` has no parameters for `path` method and always returns path with the fragment. I reported it [here](https://github.com/angular/angular/issues/27059#issuecomment-506982165).
Until it's fixed in Angular it would be great to work around it somehow in Nebular tests.
I think we can extend `SpyLocation`, override path method to respect `includeHash` parameter and provide our extended class during TextBed configuration.
You can use helper functions from `url-matching-helpers.ts`.

What do you think? 🙂
Feel free to ping me here if you need any assistance.

miqh: Ah, I see, my mistake. I obviously didn't probe far enough into the underlying cause.

PR #1845, opened 2019-07-19
chore(ci): deploy docs on master commit

### LINKED ISSUE #1531: Docs: deploy docs script
- commit to master - new docs deploy

yggg:

Closes #1531


docs/versions.json
```diff
@@ -1,12 +1,15 @@
-[
```
nnixaa: why 'json' if in both cases we use it in `ts` files?

scripts/docs/config.json
```diff
@@ -0,0 +1,4 @@
+{
```
nnixaa: same, why not `ts`?

scripts/docs/build-docs.ts
```diff
@@ -0,0 +1,117 @@
...
+  versions: Version[];
+}
+
+interface Version {
+  checkoutTarget: string;
+  name: string;
+  path: string;
+}
+
+(async function () {
```
nnixaa: so hard to read this with all the `console.log` mess... 
should we make a `log` function to at least make the code a bit more readable?

docs/versions.json
yggg: `VersionService` from the docs app also requests this file over http

scripts/docs/build-docs.ts
yggg: Done in b6a381cb62549ce30b2b0c3217b1452c3285ccf7

scripts/docs/config.json
yggg: Probably it's mine own bias to have JSON config files. Do you think it's better to change it to `config.ts`?

yggg: Fixed in 85863f0

docs/versions.json
yggg: Keeping it as json, as discussed irl, since running docs app requesting this file over the http

docs/versions.json
```diff
@@ -1,12 +1,20 @@
...
-    "path": "/nebular"
-  },
-  {
-    "version": "3.6.2",
-    "name": "3.6.2",
-    "path": "/nebular/3.6.2"
-  }
-]
+{
+  "currentVersionName": "4.1.3",
```
nnixaa: should we simply add `current`: true to the versions declaration?

yggg: done, thanks

PR #2126, opened 2019-12-09
fix: avoid to declare and export NbTable component

adwd:

NbTable is only used as base class of `NbTreeGridComponent` but is is declared and exported. Because of this, `cdk-table` cannot be used.
This PR removes `NbTable` from declare/export array to fix it.

```html
<!-- in app using Nebular -->
<table cdk-table [dataSource]="dataSource">
  ...
</table>
```

```
More than one component matched on this element.
Make sure that only one component's selector can match a given element.
Conflicting components: NbTable,CdkTable
```


yggg:
@adwd Thanks for the PR. Unfortunately, we can't keep the component decorator without adding it to the module. Instead, I changed `NbTable` component selector to a different one. It'll fix the issue as well.

PR #2286, opened 2020-03-21
Created conditional show minimize/maximize action buttons on window component

cesarochoa2006:

Adds a new feature that show/hide action buttons (minimize/maximize) with the config option "showActions" of window component that can be configured when the component is used. By default, it works like before, and the added option is by default true, so can be changed to false whenever it is desired to work. Also, added the css class no actions, to correctly display the close button when the actions buttons are not shown.
This feature is added because i had the need of hide minimize/maximize buttons of a window component.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -263,7 +263,11 @@
               "docs/favicon.ico",
               "docs/favicon.png",
               "docs/google46533d2e7a851062.html",
-              { "glob": "versions.json", "input": "docs/", "output": "/" }
+              {
+                "glob": "versions.json",
+                "input": "docs/",
+                "output": "/"
+              }
             ],
             "styles": [
               "node_modules/pace-js/templates/pace-theme-flash.tmpl.css",
@@ -326,8 +330,7 @@
             "karmaConfig": "./karma.conf.js",
             "polyfills": "docs/polyfills.ts",
             "tsConfig": "docs/tsconfig.spec.json",
-            "scripts": [
-            ],
+            "scripts": [],
             "styles": [
               "node_modules/highlight.js/styles/dracula.css",
               "docs/app/@theme/styles/styles.scss"
@@ -383,5 +386,8 @@
     "@schematics/angular:directive": {
       "prefix": "app"
     }
+  },
+  "cli": {
+    "analytics": "650ae037-17e3-4397-a8eb-7f542a78965d"
   }
 }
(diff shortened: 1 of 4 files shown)


naingaungphyo:
Any update about when will this feature be merged and available in master?

naingaungphyo:
@cesarochoa2006 
Thanks for the great work! It will be great if there is an option to disable each button that we don't want.
For me, I think think "maximize" button is not really useful. So I want to have only "minimize" button and "close" button.

cesarochoa2006:
Thanks. In fact, i have no idea of when it will be merged :(. Hopefully
soon. I have no control over it.

naingaungphyo:
fyi, 
current workaround for me to hide them manually is hiding them with overwriting css.

cesarochoa2006:
> fyi,
> current workaround for me to hide them manually is hiding them with overwriting css.

¿ But with css you hide all buttons ? I need the close button displayed

naingaungphyo:
nope, I can hide specific button with css coz they have different class to be overwritten.

cesarochoa2006:
Can hoy share with me the css that have you applied? I see no css classes
applied here

naingaungphyo:
sorry, I said wrong information. I hide with nth-child selectors like this.
`// hide maximum
nb-windows-container> nb-window > nb-card > nb-card-header> div:nth-child(2)> button:nth-child(2)
{
  display: none;
}
nb-windows-container>nb-window>nb-card>nb-card-header>.buttons
{
  justify-content: flex-end !important;
}
// hide minimize
.no-minimize-window > nb-card > nb-card-header> div:nth-child(2)> button:first-child
{
  visibility: hidden;
}`

cesarochoa2006:
Well, is a good alternative. Thanks for sharing it, i didn't know those
css.

JamesKnippel:
Would be great to be able to add a confirmation dialog on the close button as well. Shame this hasn't been pushed yet.

naingaungphyo:
@JamesKnippel  Yeah. 
Also I want to hold the state of the dialog input value before the dialog being minimized. onDestroy is called when we minimize the dialog and BOOM! all the value of inputs are gone. So even I call the dialog again from minimize state, I can't get the old form state back.

JamesKnippel:
@naingaungphyo That's my biggest problem too - I'm looking into some sort of solution with maybe storing template data on a backend with post requests and repopulating from a temporary form-data table.

I want to be able to walk away and open up the same minimized windows on another computer with the same state from old forms. I'll let you know if I come up with any workarounds. 
I really wish Nebular was more open with communications! I really love the framework and this Window is probably their most standout component with so much potential

yggg:
Closed by #2762

PR #2350, opened 2020-05-07
fix(auth): add client_id to the refresh request

alain-charles:
According to RFC4749, `client_id` has to be present in the body of token request with `refresh_token `grant_type and no client auth.

Verified on Keyclock 4.X - Invalid_credentials error if `client_id ` is not passed, even if client is not confidential (i.e. without client-secret)


alain-charles:
Hi @nnixaa
Can you please have a look at this pull request and tell me what you think about it ?
Regards
Alain

yggg (approved):
Thanks @alain-charles!

PR #2466, opened 2020-07-29
refactor(nb-icon): return null if icon wasn't found

mmfKupl:

remove icon not found error thrown so that you can dynamically change the icon, and so that after the first installation of the wrong icon, you can change it


src/framework/theme/components/icon/icon-libraries.ts
```diff
@@ -181,7 +181,7 @@ export class NbIconLibraries {
 
   protected getIconFromPack(name: string, pack: NbIconPack, shouldThrow = true): NbIcon | string {
     if (shouldThrow && !pack.icons.has(name)) {
-      throwIconNotFoundError(name, pack.name);
+      return null
     }
 
     return pack.icons.get(name);
```
yggg, line 187: Please, also remove `shouldThrow` argument as it's part of old behaviour. And then you could simplify method to:
```
if (pack.icons.has(name)) {
  return pack.icons.get(name);
}

return null;
```
And add null as possible return type.

src/framework/theme/components/icon/icon-libraries.ts
```diff
@@ -105,6 +101,10 @@ export class NbIconLibraries {
 
     const icon = this.getIconFromPack(name, iconsPack);
 
+    if (!icon) {
+      return null;
```
yggg, line 105: Add `null` to return type

src/framework/theme/components/icon/icon.component.ts
```diff
@@ -218,14 +218,17 @@ export class NbIconComponent implements NbIconConfig, OnChanges, OnInit {
   }
 
   ngOnChanges() {
-    if (this.iconDef) {
-      this.iconDef = this.renderIcon(this.icon, this.pack, this.options);
-    }
+    this.iconDef = this.renderIcon(this.icon, this.pack, this.options);
```
yggg, line 221: Not sure, but it looks like we don't need the `ngOnInit` hook if we render icon in `ngOnChanges` anyway. What do you think?

src/framework/theme/components/icon/icon.component.ts
```diff
@@ -218,14 +218,17 @@ export class NbIconComponent implements NbIconConfig, OnChanges, OnInit {
...
   }
 
   renderIcon(name: string, pack?: string, options?: { [name: string]: any }) {
     const iconDefinition = this.iconLibrary.getIcon(name, pack);
 
+    if (!iconDefinition) {
+      this.clearIcon();
+      return;
+    }
```
yggg, line 230: I found a bit confusing that calling `renderIcon` could clear it instead. Am I missing a case where we need such logic? Could we check if icon def exists before calling render and if it exists call render, otherwise call clear? Something like:
```
ngOnChanges() {
  const iconDef = this.iconLibrary.getIcon(name, pack);
  if (iconDef) {
	this.render(...);
  } else {
    this.clear();
  }
}
```

Anyway, if we keep new logic inside render, please add return types for `renderIcon` methods.

yggg: Refactored in d79b2dc

src/framework/theme/components/icon/icon.component.ts
yggg: Keeping `ngOnInit` as it will help those who create components manually (not in templates) as `ngOnChanges` won't be called for dynamic.

PR #2541, opened 2020-10-03
added support for other languages for user component 

medTL:
hello, when passing a nonlatin characters name to the user component, it doesn't show up correctly for example for Arabic: 
image
it should be like this :
/home/talel/Pictures/arabic.png
the initials should be separated and not connected


htaebi (approved):
It's good for persian and arabic languages

PR #2815, opened 2021-08-04
fix(nb-select): no error on reset with mulitple

### LINKED ISSUE #1950: Can't assign single value if select is marked as multiple when form.reset() is called
### Issue type

**I'm submitting a

* [x] bug report
* [ ] feature request

### Issue description

**Current behavior:**
NbSelect with _multiple_ property throws error **Can't assign single value if select is marked as multiple**, when `form.reset()` is triggered.

**Expected behavior:**
NbSelect should not throw any error on form.reset().

**Steps to reproduce:**

- Define a FormGroup instance and provide a property for multiple select.
   ```
  this.form = this.fb.group({
      multiproperty: [null]
   })
  ```
- Attach it to NbSelect with `multiple` component within a form.
   ```
  <nb-select multiple placeholder="Choose Options" name="multipleopts" [formControl]="form.controls['multiproperty']" fullWidth id="multiopts">
        <nb-option *ngFor="let opt of globalService.multiopts" [value]="opt.key">
              <nb-icon [icon]="opt.icon" pack="font-awesome"></nb-icon> &nbsp;
              &nbsp;{{opt.value}}
       </nb-option>
  </nb-select>
  ```
  
**Related code:**
https://stackblitz.com/edit/nbselect-multiple-issue

### Other information:

**npm, node, OS, Browser**

  Node - 10.16.0
  NPM  - 6.9.0
  OS     - Windows X
  Browser - Chrome, Firefox

**Angular, Nebular**

Angular - 8.0.0
Nebular - 4.1.2

**Error Screenshot**
image

chirichok98:

Fixes #1950 

**Before fix:**
form.reset() sets null for each control in form and null is not an array. Error was produced.

**After fix:**
When select is multiple we have additional check on null and undefined (in this case value will be empty array)

### DIFF
--- a/src/framework/theme/components/select/select.component.ts
+++ b/src/framework/theme/components/select/select.component.ts
@@ -1024,7 +1024,7 @@ export class NbSelectComponent implements OnChanges, AfterViewInit, AfterContent
 
   protected setActiveOption() {
     if (this.selectionModel.length) {
-      this.keyManager.setActiveItem(this.selectionModel[ 0 ]);
+      this.keyManager.setActiveItem(this.selectionModel[0]);
     } else {
       this.keyManager.setFirstItemActive();
     }
@@ -1156,9 +1156,14 @@ export class NbSelectComponent implements OnChanges, AfterViewInit, AfterContent
    * */
   protected setSelection(value) {
     const isArray: boolean = Array.isArray(value);
+    let formattedValue = value;
 
-    if (this.multiple && !isArray) {
-      throw new Error('Can\'t assign single value if select is marked as multiple');
+    if (this.multiple) {
+      if (value === null || value === undefined) {
+        formattedValue = [];
+      } else if (!isArray) {
+        throw new Error('Can\'t assign single value if select is marked as multiple');
+      }
     }
 
     if (!this.multiple && isArray) {
@@ -1168,10 +1173,10 @@ export class NbSelectComponent implements OnChanges, AfterViewInit, AfterContent
     const previouslySelectedOptions = this.selectionModel;
     this.selectionModel = [];
 
-    if (isArray) {
-      value.forEach(option => this.selectValue(option));
+    if (this.multiple) {
+      formattedValue.forEach(option => this.selectValue(option));
     } else {
-      this.selectValue(value);
+      this.selectValue(formattedValue);
     }
 
     // find options which were selected before and trigger deselect


src/framework/theme/components/select/select.component.ts
```diff
@@ -1156,9 +1156,14 @@ export class NbSelectComponent implements OnChanges, AfterViewInit, AfterContent
    * */
   protected setSelection(value) {
     const isArray: boolean = Array.isArray(value);
+    let formattedValue = value;
```
yggg, line 1159: Let's rename the variable to `safeValue`

src/framework/theme/components/select/select.component.ts
```diff
@@ -1156,9 +1156,14 @@ export class NbSelectComponent implements OnChanges, AfterViewInit, AfterContent
...
 
-    if (this.multiple && !isArray) {
-      throw new Error('Can\'t assign single value if select is marked as multiple');
+    if (this.multiple) {
+      if (value === null || value === undefined) {
+        formattedValue = [];
+      } else if (!isArray) {
+        throw new Error('Can\'t assign single value if select is marked as multiple');
+      }
     }
```
yggg, line 1167: Too much nesting. Let's move `null` to empty array coercion to the line right after `safeValue` declaration and add one more flag value like `isResetValue` to the `if (this.multiple && !isArray)`, something like:
```
const isResetValue = value == null;
let safeValue = value;
if (this.multiple) {
  safeValue = value ?? [];
}

if (this.multiple && !isArray && !isResetValue) {
  throw ...
```
Do you agree?

BTW, to check for both `null` and `undefined` we use loose equality comparison with `null`, it's shorter:
```value == null```

File: src/framework/theme/components/select/select.component.ts [hunk 3]
```diff
@@ -1168,10 +1173,10 @@ export class NbSelectComponent implements OnChanges, AfterViewInit, AfterContent
     const previouslySelectedOptions = this.selectionModel;
     this.selectionModel = [];
 
-    if (isArray) {
-      value.forEach(option => this.selectValue(option));
+    if (this.multiple) {
+      formattedValue.forEach(option => this.selectValue(option));
     } else {
-      this.selectValue(value);
+      this.selectValue(formattedValue);
```
yggg, line 1179: We also need to update `selectValue` method. As per documentation we treat `null` and `undefined` value as reset value, so no option should become selected when value becomes `null` or `undefined`.

yggg (changes requested):
Please, could you also check if autocomplete has the same problem? And if so, fix it as well.

chirichok98 (commented):
Comments were fixed. Added return for empty value on selectValue.

chirichok98:
> Please, could you also check if autocomplete has the same problem? And if so, fix it as well.

We have no possibility there to choose multiple items. Nothing to fix.

yggg (changes requested):

PR #2822, opened 2021-08-07
feat(nb-datepicker): change parse date func

### LINKED ISSUE #2745: nb-datepicker with [min] and not required throwing validation error when cleared
### Issue type

* [x] bug report
* [ ] feature request

### Issue description

**Current behavior:**
Setting a [min] date for the nb-datepicker on a reactive form control that is not required. Setting the date, and then clearing the date causes the form validation to be invalid with 
 { errors: 
   {  nbDatepickerMin:  { actual: Invalid Date {},  min: Fri Jan 01 2021 00:00:00 } }
}

**Expected behavior:**
Clearing the date should set the value to empty string or null (I've tried both) and still be valid upon submit.

**Steps to reproduce:**
Set the date. Then clear the date. the form control is now invalid

**Related code:**
https://stackblitz.com/edit/github-r4cprj-flbewk-79kyas?file=src%2Fapp%2Fapp.component.ts

### Other information:

**npm, node, OS, Browser**
Node: 10.14.0
OS: Windows 10
Browser: Chrome

**Angular, Nebular**
@nebular/theme: "^6.2.1"
angular: 10.2.1

chirichok98:

Fixes #2745 

What was done:
Changed parse function to handle 'Invalid Date'. In Validators we had check on `!date` but when was parsed unsuccessfully empty date equal to ''Invalid Date' string which is not `false` to be properly handled

### DIFF
--- a/src/framework/theme/components/calendar-kit/services/native-date.service.ts
+++ b/src/framework/theme/components/calendar-kit/services/native-date.service.ts
@@ -125,7 +125,9 @@ export class NbNativeDateService extends NbDateService<Date> {
    * We haven't got capability to parse date using formatting without third party libraries.
    * */
   parse(date: string, format: string): Date {
-    return new Date(Date.parse(date));
+    const parsedValue = new Date(Date.parse(date));
+    const isProperlyParsed = parsedValue && parsedValue.toString() !== 'Invalid Date';
+    return isProperlyParsed ? parsedValue : null;
   }
 
   addDay(date: Date, num: number): Date {


src/framework/theme/components/calendar-kit/services/native-date.service.ts
```diff
@@ -125,7 +125,9 @@ export class NbNativeDateService extends NbDateService<Date> {
    * We haven't got capability to parse date using formatting without third party libraries.
    * */
   parse(date: string, format: string): Date {
-    return new Date(Date.parse(date));
+    const parsedValue = new Date(Date.parse(date));
+    const isProperlyParsed = parsedValue && parsedValue.toString() !== 'Invalid Date';
+    return isProperlyParsed ? parsedValue : null;
```
yggg, line 130: It's a breaking change. Users could rely on `parse` always returning a `Date` object. With updated behavior, there is possibly a `null.someDateMember` call. What do you think about checking for invalid dates in the validators themselves?
To check for invalid date, let's better use `isNaN(date.getTime)`. We could add the `isValid` method to the `NbDateService`. Don't forget to update moment implementation (NbMomentDateService) and this method (https://github.com/akveo/nebular/blob/7753027cb9e1efec28cc26a54b18a53b5f2e1eb0/src/framework/theme/components/calendar-kit/services/native-date.service.ts#L68).

Also, I'm concerned about the correctness of min/max validators for range pickers.
https://github.com/akveo/nebular/blob/7753027cb9e1efec28cc26a54b18a53b5f2e1eb0/src/framework/theme/components/datepicker/datepicker.directive.ts#L394
https://github.com/akveo/nebular/blob/7753027cb9e1efec28cc26a54b18a53b5f2e1eb0/src/framework/theme/components/datepicker/datepicker.directive.ts#L404
`datepickerAdapter.parse` on range picker would return a `{ start: D, end: D }` objects, and we use it directly in the `dateService.compareDates` which expects Date objects. Could you please check does it works as expected?

And the last thing, let's change ternary operators in the validators to `if` to make it readable.

yggg (changes requested):

PR #3022, opened 2022-02-06
fix: select with multiple option NgModel don't fill the component

augustoicaro:

Fix select component with multiple selections not been filled when using NgModel.


augustoicaro:
I tried to reproduce the problem in ngx-admin, but I could not reproduce the problem that I have in my project using the ngx-admin as a template. So I closed because it is a project-related problem.

PR #3028, opened 2022-02-15
feat(auth): added firebase apple strategy

lukas-mertens:

I added a firebase apple strategy to allow authentication by apple id. This is crucial for platforms which feature an ios app, as apple requires app developers to allow people to use "sign in with apple" if other social logins are allowed as well. Therefore this is an important feature.

As it is not trivial to add the required firebase-settings to the current playground, this MR doesn't feature a playground/unit-tests yet. If you think I should work on that, please point me into the right direction, as the other strategies don't have unit-tests either, as far as I could see.

### DIFF
--- a/docs/articles/auth/firebase-authentication.md
+++ b/docs/articles/auth/firebase-authentication.md
@@ -2,10 +2,12 @@
 
 `@nebular/firebase-auth` allows authentication in firebase applications with `@nebular/auth`.
 The package provides the following strategies:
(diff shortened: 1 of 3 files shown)


lukas-mertens:
@yggg or some other maintainer would have to review and approve my change. Is an apple firebase auth strategy wished? If that is the case, I would rebase so that this can be merged

lukas-mertens:
As no reviewer seems to be available, I will close the PR

PR #2827, opened 2021-08-13
fix(tabset): lazy loading

evtkhvch:
### Please read and mark the following check list before creating a pull request:

 - [x] I read and followed the [CONTRIBUTING.md](https://github.com/akveo/nebular/blob/master/CONTRIBUTING.md) guide.
 - [x] I read and followed the [New Feature Checklist](https://github.com/akveo/nebular/blob/master/DEV_DOCS.md#new-feature-checklist) guide.
 
 #### Short description of what this resolves:

Closes #1827
Closes #1820

- add `nbTabContent` directive
- `lazyLoad` property in `NbTabComponent` is deprecated
- `init` property in `NbTabComponent` is deprecated

how to use:
```
<nb-tab tabTitle="title">
  <ng-template nbTabContent>
    ...content
  </ng-template>
</nb-tab>
```

Issue #1827, opened 2019-07-17
Fix lazy loading in Tab component


ESadouski:
### Issue type

**I'm submitting a ...**  (check one with "x")

* [x] bug report
* [ ] feature request

### Issue description

**Current behavior:**
Lazy loading doesn't work. Content loaded even if it's not active

**Expected behavior:**
When tab marked as lazy-loaded it should be initialized only when its active 

**Steps to reproduce:**
Mark tab as lazy=true

**Related code:**

```
  @Input()
  set lazyLoad(val: boolean) {
    this.init = convertToBoolProperty(val);
  }
```

### Other information:

**npm, node, OS, Browser**
```
<!--
Node, npm: `node --version` and `npm --version`
OS: Windows (7/8/10). Linux (incl. distribution). macOS (El Capitan? Sierra?)
Browser: Chrome/Safari/Firefox/etc?
-->
```

**Angular, Nebular**
```
<!--
Check your `package-lock.json` or locate a `package.json` in the `node_modules` folder.
-->
```

ESadouski:
For those who faced this problem, `route-tabset` can be used as a temporary solution. Or please read https://github.com/angular/angular/issues/13921#issuecomment-272660678

maihannijat:
Duplicate of #1820

redriverit:
Any Update on this? I am seeing all content within my tabs load instantly and not when the tab is active. As mentioned above [lazyLoad] does not change the functionality:
<nb-tab tabTitle="Parties" [lazyLoad]="false">

dmk1111:
Hi, guys! Any update on this? For now, with `[lazyLoad]="false"` content is not rendered in the template but all hooks (calls to backend, etc) in child components are still executed. Is there a way to restrict this completely until tab is selected first time? This is pretty annoying in case we have a lot of tabs.

Issue #1820, opened 2019-07-16
nb-tabset lazyLoad and (changeTab) not working


maihannijat:
### Issue type

**I'm submitting a ...**  (check one with "x")

* [x ] bug report
* [ ] feature request

### Issue description
I have `nb-tabset` with two tabs and one component in each tab. I want to lazy load the second component. The second component makes API call and I want to make that only if user clicks on the tab or the tab become active. The `[lazyLoad]="{{true}}"` or `lazyLoad`  decorators do not work. Tried `(changeTab)` event or `(click)` but these do not fire either.

**Current behavior:**
`lazyLoad` loads component instantly. 
`(changeTab)` and `click` do not fire.

**Expected behavior:**
`lazyLoad` loads content when the tab is active and `(changeTab)` and `(click)` events should fire.

**Related code:**
```
<nb-tab tabTitle="Map">
  <div class="row">
    <div class="col-md-6">
      <app-field-info></app-field-info>
      <app-field-forecast></app-field-forecast>
    </div>
    <div class="col-md-6">
      <app-map></app-map>
    </div>
  </div>
</nb-tab>
<nb-tab tabTitle="Calendar" (changeTab)="displayCalendar()" lazyLoad>
  <app-calendar *ngIf="isCalendarActive"></app-calendar>
</nb-tab>
</nb-tabset>
```

### Other information:

**Angular, Nebular**
```
Angular 8 and Nebular 4.1.2
```

rogrdat:
Any plans to address this issue?

alexandr-efimov:
looks like it is still actual :(

PR #3049, opened 2022-04-06
fix(tabset): Use the init state to determine if the nb-template should be loaded

Liberatys:
fix(tabset): Use the init state to determine if the nb-template should be loaded

Use the existing state indication to prevent the initial rendering of templates with the lazy directive

Ref: #2827 
Ref: #1827 
Ref: #1820 

```
<nb-tab tabTitle="title">
  <ng-template nbTabContent>
    ...content
  </ng-template>
</nb-tab>
```


Liberatys:
@yggg

PR #3091, opened 2022-08-02
feat(select): options search

sashaqred:

### DIFF
--- a/src/framework/theme/components/select/select.component.html
+++ b/src/framework/theme/components/select/select.component.html
@@ -1,4 +1,5 @@
 <button
+  [hidden]="isOptionSearchInputAllowed"
   [disabled]="disabled"
   [ngClass]="selectButtonClasses"
   (blur)="trySetTouched()"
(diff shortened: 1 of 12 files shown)


src/framework/theme/components/select/select.component.html
```diff
@@ -1,4 +1,5 @@
 <button
+  [hidden]="isOptionSearchInputAllowed"
```
Tibing, line 2: why is it hidden? why not *ngIf?

src/framework/theme/components/select/select.component.ts
```diff
@@ -849,7 +862,7 @@ export class NbSelectComponent
       return this.selectionModel.map((option: NbOptionComponent) => option.content).join(', ');
     }
 
-    return this.selectionModel[0].content;
+    return this.selectionModel[0]?.content;
```
Tibing, line 865: why do we need this change?

src/framework/theme/components/select/select.component.ts
```diff
@@ -927,6 +950,10 @@ export class NbSelectComponent
     if (this.isOpen) {
       this.ref.detach();
       this.cd.markForCheck();
+      this.selectClose.emit();
+
+      this.optionSearchInput.nativeElement.value = this.selectionView;
```
Tibing, line 955: why don't we use FormControl for optionSearchInput?

Tibing (changes requested):

Tibing:
Also, let's add tests

src/framework/theme/components/select/select.component.ts
sashaqred: Earlier `selectionView` variable was used in the button only when something was selected. Now it is used in `optionSearchInput` all the time

src/framework/theme/components/select/select.component.html
sashaqred: Just simpler. Nebular subscribes to some button events like focus. And with `*ngIf` we should implement unsubscription and resubscription mechanisms.  Looks too complicated for now

src/framework/theme/components/select/select.component.ts
sashaqred: `FormControl` adds some complexity to updating search input value when something in select is updated. Probably should be improved, but for now, I think it's not bad.

Tibing (approved):

Tibing (approved):

